PR #6, opened 2015-08-08
Begin Phoenix integration.

alanpeabody:


alanpeabody:
@bcardarella @hunterboerner I would like to improve the experience of developers using JaSerializer in Phoenix apps and would appreciate some insight from you guys.

In playing around it seemed to me that just extending the built in view to be the serializer was the cleanest integration, please see the usage example in this PR and let me know if you agree.



Other things that I have coming (in later PRs):

* Determining the default "type" value by reflecting on the module name.
* Deprecation of `serializer/2` macro - `attributes` and `has_many`/`has_one` will be available outside it.
* Automatic, but overridable, handling of Ecto relationships in has_many/has_one.
* Filtering of attributes based on conn (current_user).
* Sparse fieldset support
* Request time relationship inclusion

Thanks for your time and insights!

hunterboerner:
Looks good so far! Can you explain how you would implement these points?

- Deprecation of serializer/2 macro - attributes and has_many/has_one will be available outside it.
- Filtering of attributes based on conn (current_user).
- Sparse fieldset support
- Request time relationship inclusion

alanpeabody:
The latter three are a bit up in the air, but:

Deprecation of `serializer` macro can be illustrated best by the tests for the implementation itself: https://github.com/AgilionApps/ja_serializer/compare/dsl-tweaks#diff-559e2a3e4df6aeefe57a1a6f4ce66881R1

Filtering attributes may end up with an optional `filter_attributes/2` callback. I think it will probably also just return an array of fields, but I may instead opt for the ability to override `attributes/2` which is expected to return the actual data.

Sparse fieldsets and request time includes will be options you pass to .format (or render) and get from the params.

hunterboerner:
> Sparse fieldsets and request time includes will be options you pass to .format (or render) and get from the params.

What do you mean by request time? Performance info about how long it took to process your request?

alanpeabody:
I mean: http://jsonapi.org/format/#fetching-sparse-fieldsets and http://jsonapi.org/format/#fetching-includes

hunterboerner:
Oh, that makes much more sense. Though being able to respond with performance metric would be awesome. Going to go bug someone about that. :)

bcardarella:
Sorry I'm catching up.

I'm :+1: on doing through the View. I haven't been making heavy use of Views in Phoenix as I've just gone directly to `json/2` in the Controller but as I understand it Views are the preferred method for this stuff.

Also, it might be outside the scope of this PR, or even this library, but it might be good to discuss how best to go about *receiving* json-api payloads from the client into Phoenix. Ember Data, for example, will make requests to the backend with json-api data sets.

I have a simple enough serializer that I run https://dockyard.com/blog/2015/08/05/building-json-api-endpoints-with-phoenix but I suspect I'll need a better abstraction with more complex payloads (i.e. relationships)

alanpeabody:
@bcardarella Yes, receiving payloads is an interesting problem. In [relax](https://github.com/AgilionApps/relax/blob/master/lib/relax/plug_parser.ex) I used a similar plug to what you have defined, but your config/mimetype approach is much better. I would not be opposed to adding a JaSerializer.Deserializer plug that does what your blog post mentioned.

I also think some sort of whitelisting/normalization of attributes into a single map at the controller/action level would be useful. Basically a Dict.take + merge of attributes and relationships, while also appending `_id` to relationship fields.

```elixir
JaSerializer.permit_params(conn.params, attributes: [:title, :body], relationships: [:category, :author])
# %{"title" => "...", "body" => "...", "category_id" => "...", "author_id" => "..."}
```

PR #7, opened 2015-08-09
Small changes to DSL.

alanpeabody:
* `type/0` callback now expected to return serializer type.
* Default `type/0` implementation guesses at type based on module name.
* `serializer/2` macro deprecated in favor of `type/1`.
* `id/2` callback added and documented, expected to return id.
* `attributes/2` callback added, expected to return actual map of attributes to be returned.
* Default `attributes/2` implementation finds values as populated by `attributes/1`.

//cc @beerlington Some of the changes we talked about are here.

### DIFF
--- a/lib/ja_serializer/serializer.ex
+++ b/lib/ja_serializer/serializer.ex
@@ -75,6 +75,52 @@ defmodule JaSerializer.Serializer do
   """
   defcallback type() :: String.t
 
+  @doc """
+  Returns a map of attributes to be mapped.
+
+  The default implimentation relies on the `attributes/1` macro to define
+  which fields to be included in the map.
+
+      defmodule UserSerializer do
+        attributes [:email, :name, :is_admin]
+      end
+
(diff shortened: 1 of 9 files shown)


lib/ja_serializer/serializer.ex
```diff
@@ -42,61 +38,157 @@ defmodule JaSerializer.Serializer do
...
+  @doc false
+  def default_attributes(serializer, model, conn) do
+    Enum.reduce serializer.__attributes, %{}, fn(attr, acc) ->
+      Map.put(acc, attr, apply(serializer, attr, [model, conn]))
+    end
+  end
+
+  @doc false
+  defmacro serialize(type, do: block) do
+    IO.puts "serialize/2 is deprecated, please use `type/1` instead"
```
hunterboerner: `type` is of arity 0, not 1.

alanpeabody: Good catch, thanks.

PR #24, opened 2015-10-03
Implement pagination links

vysakh0:
As per the [json api spec on pagination links](http://jsonapi.org/format/#fetching-pagination), make a top level `links` object with keys

- self
- first
- last
- prev
- next

(P.S I could not run tests without removing `only: :test` option in `mix.exs` file for `Poison` library. )

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,25 @@ model
 |> Poison.encode!
 ```
 
+#### Pagination Usage
+
+Pass a `page` map with keys `page_number`, `total_pages` and `page_size` to the format method.
+
+```elixir
+model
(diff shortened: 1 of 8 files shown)


alanpeabody:
@vysakh0 this is a great start, I really appreciate the contribution!

I would like to think a bit more about a couple things, I have pulled this commit into a branch (https://github.com/AgilionApps/ja_serializer/compare/pagination) to play with tomorrow on a long plane trip.

In particular I want to think about/play with:

* Is there a better way to deal with singular vs list `location` paths? (Can/should this come from the conn?)
* Which interfaces should enforce the JSON-API params (first, last, next, prev) vs the scrivener params (page, page_size, etc).

alanpeabody:
Merged in #25

PR #25, opened 2015-10-11
Pagination

alanpeabody:
@vysakh0, I have built upon what you started and made a few tweaks:

1) Stronger opinions when using Scrivener.Page. (Just pass it in as the model)
2) Weaker opinions on pagination in general. (Allow cursor based etc to be custom implemented)
3) Path/URL is now based on the current request's path.
4) Params are now in the JSON-API spec suggested `page` namespace.

Let me know if you have any feedback otherwise I will merge this up and release a new version.


vysakh0:
@alanpeabody Cool :+1:

Issue #26, opened 2015-10-14
Incorrect representation of empty has_one/has_many relationships


dmarkow:
With a `has_one :author` relationship, if there is no linked author, you get this:

```json
"relationships": {
  "author": {
    "data": {
      "type": "author",
      "id": ""
    }
  }
}
```

However, the JSON API 1.0 spec says that resource linkage for an empty to-one relationship **must** be `null`, so I would expect:

```json
"relationships": {
  "author": {
    "data": null
  }
}
```

A `has_many :authors` scenario is slightly different. Right now, an empty `has_many` relationship gives you:

```json
"relationships": {
  "authors": {}
}
```

whereas the JSON API spec requires an empty array for empty to-many relationships:

```json
"relationships": {
  "authors": {
    "data": []
  }
}
```

If I have time this week I can start working on a PR, but wanted to raise it as an issue in the mean time. Thanks!

dmarkow:
For has_many relationships, the cause was easily found. The `Utils` module's `put_if_present` function ignores both `nil` and `[]` and just returns the untouched `dict` object. In the relationship formatter, since calling `put_if_present` on  `:links` and `:data` does nothing to the dict, an empty dict is returned. Still not sure what the answer should be at this point, though.

For the has_one issue, we might be able to add a function clause in `JaSerializer.Builder.ResourceIdentifier` for `build` to handle the case where `id` is `nil`:

```elixir
def build(nil, _type), do: nil
```
However, this will then make it do the same thing as an empty has_many...

dmarkow:
One possible approach might be to use `:null_value` and `:empty_list` atoms to help signal that we want to actually add explicit `null` and `[]` entries to the formatted output. We would add `put_if_present` clauses to actually add the null/empty values to `dict` if those atoms are used (while still ignoring actual `nil` and `[]` values):

```elixir
def put_if_present(dict, key, :null_value), do: Dict.put(dict, key, nil)
def put_if_present(dict, key, :empty_list), do: Dict.put(dict, key, [])
```

Then a few updates to `JaSerializer.Builder.ResourceIdentifier`. For has_many, add a case pattern for an empty array, and for has_one, a new function clause to handle a blank `id`:

```elixir
def build(%{serializer: serializer} = context, type, name) do
  serializer
  |> apply(name, [context.model, context.conn])
  |> case do
    [] -> :empty_list
    many when is_list(many) -> Enum.map(many, &build(&1, type))
    one -> build(one, type)
  end
[...]

PR #29, opened 2015-10-15
Fix empty relationships

dmarkow:
Here's my first attempt at fixing #26, though it's definitely not ready to merge yet.

I agree on the integration test, so I started with that. I added `has_many :likes` and `has_one :excerpt` and set them both to be empty. As expected, the test fails.

To handle a blank `has_one`, I updated the `ResourceIdentifier` formatter to format an empty relationship as `nil` (instead of a resource identifier object with a blank `id` field, e.g. ```{"type": "excerpt", "id"; ""}```).

Next, I updated `put_if_present` in the `Utils` module to explicitly add `nil`/ `[]` to `dict` only when it's for the `:data` key. However:

1. While the JSON API spec only requires explicit `null`/`[]` values when dealing with the `data` object, I don't really like explicitly matching on the `:data` key. But maybe I'm just overthinking it.
2. More importantly, because of this approach (specifically, the guard clause on the `nil` case), `{"data": ...}` will be added to **every** relationship in the output, even if we didn't ask it to (e.g. we may have just asked for `has_one :excerpt, link: "..."`). And even worse, if we didn't ask for `:data`, it'll be set to `null` even if there really are items in the relationship.

So now I'm stuck. The `JaSerializer.Builder.Relationship` struct includes the `:data` key, which defaults to `nil`. If we have an empty has_one relationship, that also sets `:data` to nil. So how do we tell the difference between "it's nil because we never set it to anything" vs "it's nil because that's specifically what we want"? (This is why I originally suggested using a `:null_value` atom, though I agree that doesn't feel right).

### DIFF
--- a/lib/ja_serializer/builder/resource_identifier.ex
+++ b/lib/ja_serializer/builder/resource_identifier.ex
@@ -7,6 +7,8 @@ defmodule JaSerializer.Builder.ResourceIdentifier do
     serializer
     |> apply(name, [context.model, context.conn])
     |> case do
+      [] -> [:empty_relationship]
+      nil -> :empty_relationship
       many when is_list(many) -> Enum.map(many, &build(&1, type))
       one -> build(one, type)
     end
--- a/lib/ja_serializer/formatter/relationship.ex
+++ b/lib/ja_serializer/formatter/relationship.ex
@@ -3,7 +3,7 @@ defimpl JaSerializer.Formatter, for: JaSerializer.Builder.Relationship do
 
   def format(rel) do
     json = %{}
-    |> Utils.put_if_present(:data, JaSerializer.Formatter.format(rel.data))
+    |> Utils.add_data_if_present(JaSerializer.Formatter.format(rel.data))
     |> Utils.put_if_present(:links, Utils.array_to_hash(rel.links))
     {Utils.format_key(rel.name), json}
   end
--- a/lib/ja_serializer/formatter/utils.ex
+++ b/lib/ja_serializer/formatter/utils.ex
@@ -9,6 +9,11 @@ defmodule JaSerializer.Formatter.Utils do
   def put_if_present(dict, key, val), do: Dict.put(dict, key, val)
 
   @doc false
+  def add_data_if_present(dict, :empty_relationship), do: Dict.put(dict, :data, nil)
+  def add_data_if_present(dict, [:empty_relationship]), do: Dict.put(dict, :data, [])
+  def add_data_if_present(dict, val), do: put_if_present(dict, :data, val)
+
+  @doc false
   def array_to_hash(nil),   do: nil
   def array_to_hash([nil]), do: nil
   def array_to_hash(structs) do
--- a/test/ja_serializer/json_api_spec/compound_document_test.exs
+++ b/test/ja_serializer/json_api_spec/compound_document_test.exs
@@ -28,6 +28,18 @@ defmodule JaSerializer.JsonApiSpec.CompoundDocumentTest do
             { "type": "comments", "id": "5" },
             { "type": "comments", "id": "12" }
           ]
+        },
+        "likes": {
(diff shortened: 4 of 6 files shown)


lib/ja_serializer/formatter/resource_identifier.ex
```diff
@@ -1,4 +1,5 @@
 defimpl JaSerializer.Formatter, for: JaSerializer.Builder.ResourceIdentifier do
+  def format(%{id: nil}), do: nil
```
alanpeabody: :+1:

alanpeabody:
Thanks! I think this looks good. I will pull it down today/tonight, see if something strikes me that would help with the issues you raised (particularly the 2nd one), and get back to you with feedback.

alanpeabody:
Okay, so after tracing through this I have actually come back to either using your option special atom option or extending the struct to contain more context (eg `%Relationship{includes_data: true}`).

I actually think with a couple tweaks the atom option is pretty good:

My initial reaction was that `:null_value` felt wrong because it seems so close to the existing `:nil` atom. I think however if we change the name to be more representative it "feels" better. Something like `:not_found` and maybe even `[:not_found]`.

I think this could be added in easily to the case statement in JaSerializer.Builder.ResourceIdentifier you identified in #26 .

```elixir
[] -> [:not_found]
nil -> :not_found
```

The other thing I would probably consider is not using Utils.put_if_present for the `:data` in the Relationship formatter. I think a custom `add_data_if_present` or something similar that isolates the `:not_found` handling logic would help reveal the intent of the code.

I just played briefly with this, so I may have overlooked something.

Thanks for digging into this and contributing!

dmarkow:
Nice, that worked out perfectly:

- Added `:not_found` and `[:not_found]` to the builder (Note: I might be more inclined to use the atom `:empty_relationship` since that's inline with how JSON API refers to it?)
- Got rid of the `def format(%{id: nil}), do: nil` line I originally added in the `ResourceIdentifier` formatter since we don't need it anymore (there will no longer be a situation where `id` is `nil`)
- Added a new `add_data_if_present` that explicitly adds `nil`/`[]` for `:not_found`/`[:not_found]`, otherwise it defers to `put_if_present`
- Also updated the tests to make sure we're not including `data: null` in the formatted output when we never asked for `:data` in the first place.

I think this now meets all of the requirements for dealing with empty relationships. If everything looks good to you, I can squash these commits so it can be merged. (If you prefer pull requests include updates to the changelog, I can include that too). Thanks!!!

alanpeabody:
This is great, thank you!

I can update the change log when I push out a new release.

If you want to change the atom to `:empty_relationship` I agree that is better then not found.

I will merge this up today either way and push out a release this weekend (I want to hit the other bug you addressed and spend a little more focused time with `mix profile.fprof` to make sure there are no terrible bottlenecks in the code).

dmarkow:
Updated and squashed. I really appreciate your time helping me get this working. I'm very new to Elixir (coming from a long-standing Ruby mindset) and I feel like working through this helped me learn a bunch about pattern matching, protocols, etc.

alanpeabody:
Awesome, I have also learned a lot building this library.  This has been my first use of protocols, and I may have admittedly gone a little over the top with macros to begin with but it has been really interesting.

I would be interesting in your feedback on the code in general, as well as how it is working for you as a library.

dmarkow:
Haven't had much of a chance to use it so far -- it's my first time using the JSON API format, so I'm not even at the point that I have a client ready to consume my API, and I ran into the empty relationship issue early on. But the serializer API is very familiar coming from ActiveModelSerializers in Ruby, so it was easy to dive right in. And having absurdly fast response times helps too.

I'll probably have some more time next week to work on my project, so I'll have more feedback soon (and depending on how complex my API gets, I may be interested in the sparse fieldset/relationship stuff and might be able to help get that going)

Issue #12, opened 2015-08-12
Specify relationships to be included.


alanpeabody:
http://jsonapi.org/format/#fetching-includes

likely means a new relationship key such as `optional_include`.

green-arrow:
Do you have any thoughts on how you want this to work? After looking over things, my thought was to pass the comma separated list of includes directly to the `format` method of the serializer, and take things from there.

alanpeabody:
I had started on this a while ago: https://github.com/AgilionApps/ja_serializer/compare/include-param

I haven't had an opportunity to get back to it, but here are some thoughts if you want to pick this up:

I think it makes sense to pass includes in as an option to the format method. I am not sure if it should be passed in as the raw string or as pre-parsed options.

2nd level includes are much more complicated (see the `comments.author` in the [example](http://jsonapi.org/format/#fetching-includes)). I am not particularly proud or happy with the existing include logic, and I foresee this making it more complicated.

The sparse attributes (fields) param should end up with an extremely similar API and some similar implementation details.

PR #31, opened 2015-10-20
Adding support for relationships with custom ids.

green-arrow:
When a related resource has a custom id def in its serializer, that custom id should be used when specifying resource identifiers that go under the relationships data property.

**Edit:** Changed resource *objects* to resource *identifiers*

### DIFF
--- a/lib/ja_serializer/builder/relationship.ex
+++ b/lib/ja_serializer/builder/relationship.ex
@@ -29,6 +29,7 @@ defmodule JaSerializer.Builder.Relationship do
     |> case do
       nil  -> relation
       type ->
+        context = Map.put(context, :resource_serializer, opts[:include])
         Map.put(relation, :data, ResourceIdentifier.build(context, type, name))
     end
   end
--- a/lib/ja_serializer/builder/resource_identifier.ex
+++ b/lib/ja_serializer/builder/resource_identifier.ex
@@ -9,16 +9,26 @@ defmodule JaSerializer.Builder.ResourceIdentifier do
     |> case do
       [] -> [:empty_relationship]
       nil -> :empty_relationship
-      many when is_list(many) -> Enum.map(many, &build(&1, type))
-      one -> build(one, type)
+      many when is_list(many) -> Enum.map(many, &build(&1, type, context))
+      one -> build(one, type, context)
     end
   end
 
+  def build(%{} = model, type, %{resource_serializer: resource_serializer} = context) do
+    id = cond do
+      resource_serializer ->
+        apply(resource_serializer, :id, [model, context.conn])
+      true ->
+        Map.get(model, :id)
+    end
+
+    build(id, type)
+  end
+
   def build(%{} = model, type) do
-    %__MODULE__{
-      type: type,
-      id: Map.get(model, :id)
-    }
+    id = Map.get(model, :id)
+
+    build(id, type)
   end
 
   def build(id, type) do
--- a/test/ja_serializer/builder/relationship_test.exs
+++ b/test/ja_serializer/builder/relationship_test.exs
@@ -1,0 +1,48 @@
+defmodule JaSerializer.Builder.RelationshipTest do
+  use ExUnit.Case
+
+  defmodule ArticleSerializer do
+    use JaSerializer
+
+    def type, do: "articles"
+    attributes [:title]
+    has_many :comments,
+      include: JaSerializer.Builder.RelationshipTest.CommentSerializer
+  end
+
+  defmodule CommentSerializer do
+    use JaSerializer
+    def id(model, _conn), do: model.comment_id
+    def type, do: "comments"
+    location "/comments/:id"
+    attributes [:body]
+  end
+
+  test "custom id def respected in relationship data" do
+    c1 = %TestModel.CustomIdComment{comment_id: "c1", body: "c1"}
+    c2 = %TestModel.CustomIdComment{comment_id: "c2", body: "c2"}
+    a1 = %TestModel.Article{id: "a1", title: "a1", comments: [c1, c2]}
+
+    context = %{model: a1, conn: %{}, serializer: ArticleSerializer, opts: []}
+    primary_resource = JaSerializer.Builder.ResourceObject.build(context)
+
+    %JaSerializer.Builder.ResourceObject{
+      relationships: [%JaSerializer.Builder.Relationship{:data => rel_data}]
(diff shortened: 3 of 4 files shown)


alanpeabody:
Thanks for your contribution @green-arrow!

I would like to take a bit of time to think about this.

I understand the custom id use case, and I would like the solution to work for resource-identifiers that are not side-loaded/included as well. That said, it does seem like it makes sense to use the serializer's id if a serializer is included.

green-arrow:
@alanpeabody I'm happy to contribute!

Could you elaborate on what you mean by wanting the solution to work for resource identifiers that are not side-loaded / included? After looking at the spec, it seems that the two cases where resource identifiers are used are:

1) In primary data
2) In a relationship within a resource object

Is one of the goals of this library to support both points 1 and 2? As of now, I don't think that point 1 is possible. If supporting that is a goal, then some modifications to this PR will definitely be needed.

As for point 2, I believe this PR covers this scenario entirely. The two use cases for this point would be:

- User specifies an array of `id`s - In this case no serializer lookup is needed, the user would need to map the correct `id`s manually
- User specifies a model / array of models to be included - this is the use case handled by this PR.

So I guess what it all comes down to is: do we want to take into account resource identifiers being allowed in the primary data portion of the response?

As an aside, this seems like a potentially good opportunity to separate out the `include` property on the relationship into something like `serializer` instead. This would allow users to specify a serializer to be used when creating resource identifiers / objects for this relationship, while preparing us to take on #12 (specify relationships to be included). 

`included` would become something like a boolean. If `true`, the resource would always be included in the serialized payload. `optional_include` could also become a boolean, and if `true` would require a lookup somewhere else to detect whether the relationship should be side-loaded or not. The logic for side-loading a resource would then become something like:

```
if include or (optional_include and user specified this relationship should be included) do
  side-load this relationship
end
```

Sorry for the wall of text! I'm happy to pitch in with additional PRs / additional conversations to try and flesh these ideas out :smile:

green-arrow:
I did a quick test on separating out the serializer definition for the relationship from the `include` option, and made `include` a boolean. It turned out to be a fairly easy refactor, and I think sets things up nicely for specifying optional includes.

[Compare to master](https://github.com/AgilionApps/ja_serializer/compare/master...green-arrow:specify_serializer_for_relationship)

[Branch](https://github.com/green-arrow/ja_serializer/tree/specify_serializer_for_relationship)

alanpeabody:
Thanks for the further details! I know care and thought goes into every pull request so I like to try to put thought in when reviewing as well.

I am going to try to clarify my original comment first. Here is how I read the spec and how the existing implementation is working:

1) Included relationships

```elixir
has_many :comments, included: CommentSerializer

def comments
  [%Comment{id: 1}, %Comment{id: 2}]
end
```

Results in a data structure like this (please excuse my poorly formatted json):

```javascript
{
  data: {
     id: /*..*/,
     attributes: {/*..*/},
     relationships: {
      comments: [
        {id: 1, type: "comment"}, {id: 2, type: "comment"}
      ]
    }
  },
  includes: [
    /* actual serialized comments */
  ]
}
```

2) Referenced, but not included relationships

```elixir
has_many :comments, type: "comment"

def comments
  [%Comment{id: 1}, %Comment{id: 2}]
end
```

Which results in the same thing, but minus the includes key:

```javascript
{
  data: {
     id: /*..*/,
     attributes: {/*..*/},
     relationships: {
      comments: [
        {id: 1, type: "comment"}, {id: 2, type: "comment"}
      ]
    }
  }
}
```

Some things of note are that the `type` comes from the serializer in the first case, but could be passed in by the `type` option as in the second example instead.

In both cases I do think we need to support custom ids.

Your PR supports custom ids for included relationships nicely.

What I was wondering is if we should support custom ids via the DSL for both included and non-included relationships, or if just something like this is sufficient (which I believe you mentioned):

```elixir
has_many :comments, type: "comment"

def comments
  [%Comment{cid: 1}, %{Comment{cid: 2}] |> Enum.map(&(&1.cid))
end
```

green-arrow:
Thanks for the clarification!

Here's my thought for the two items you pointed out:

1) Included relationship

```elixir
has_many :comments, serializer: CommentSerializer, included: true

def comments
  [%Comment{id: 1}, %Comment{id: 2}]
end
```

The above would result in the same JSON you posted for item 1. The other branch I made (https://github.com/green-arrow/ja_serializer/tree/specify_serializer_for_relationship) is where I started down this road.

2) Referenced, but not included relationships

```elixir
has_many :comments, type: "comment"

def comments
  [%Comment{id: 1}, %Comment{id: 2}] |> Enum.map(&(&1.is))
end
```

This is where I'm wondering if we support specifying reference, not included relationships as a model / array of models without also specifying a serializer. My gut reaction would be that, if you want to specify a model / array of models, you also must include a serializer. If a serializer is not included, the relationship should be an already mapped list of `id`s.

alanpeabody:
RE: #12 && `serializer` vs `included`

I have been sitting on this and thinking about it quite a bit. I did an initial spike using `default: true/false` a few weeks ago (https://github.com/AgilionApps/ja_serializer/compare/include-param) but a couple of things didn't sit right with me. In particular the key names include and default didn't make a ton of sense to me, so I do like the change to `serializer`.


I really like your proposal/last comment:

```elixir
has_many :comments, serializer: CommentSerializer, included: true
```

The idea of allowing the type and id in a separate serializer is really nice.

My big concern is breaking the API from include to serializer.

Let me sleep on this, but I think I will probably want to go ahead with renaming (your 2nd branch), maybe we can figure out a nice way to deprecate it.

Thanks!

green-arrow:
Sounds like a plan. I appreciate the discussion and all the thought you've put into this review and this library!

alanpeabody:
Okay, so here is what I am thinking:

1) I want to include this change.
2) I also want to include the `:include` -> `:serialize` change.
3) I would like the existing `include` config if passed a module to output a deprecation message and be converted to the equivalent of `:serialize` with `:include` true.

If you have the time and inclination to do the last that is great, if not I can pull down what you have and address the last myself in the next few days.

green-arrow:
Sounds like a plan to me. How would you like to proceed? 

We could merge this PR and I could make the deprecation change you requested in my other branch then make an additional PR. 

Or I can just roll them all into one PR and close this one.

alanpeabody:
@green-arrow I merged this as it stands alone, but I want to nail the other 2 before pushing another release. Thanks!

PR #33, opened 2015-10-22
Specify serializer for relationship

green-arrow:
Moving `:include` option to new `:serializer` option. `:include` option becomes a boolean.

See #31 for additional details.

### DIFF
--- a/lib/ja_serializer/builder/included.ex
+++ b/lib/ja_serializer/builder/included.ex
@@ -35,6 +35,9 @@ defmodule JaSerializer.Builder.Included do
   # Find relationships that should be included.
   defp relationships_with_include(context) do
     context.serializer.__relations
+    # This filter is where we would test for opts[:include]
+    # OR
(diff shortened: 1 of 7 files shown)


lib/ja_serializer/builder/included.ex
```diff
@@ -55,4 +62,23 @@ defmodule JaSerializer.Builder.Included do
   defp reject_known(resources, included, primary) do
     Enum.reject(resources, &(&1 in included || &1 in primary))
   end
+
+  defp normalize_relation_opts(opts, trigger_deprecation \\ false) do
```
alanpeabody: Can we move the deprecation and normalization to compile time in the `Serializer.has_many/2` and `Serializer.has_one/2` macros?

green-arrow: Fixed with commit fe268f0f2d5c2703f7337f96fae569402ec67a9d

alanpeabody:
Awesome, thanks!

PR #39, opened 2015-11-19
Getting optional includes to work

### LINKED ISSUE #12: Specify relationships to be included.

green-arrow:
A properly formatted 'include' string, which adheres to the JSONApi spec, can be passed as an additional option to the 'format' method on a serializer, and only the resources specified are included.

I went the route of just passing down the comma separated 'include' string straight to `format`, that way we minimize the amount of work that the consumer has to do in their controller. I know you were on the fence about that one so I just picked one and ran with it :)

I added tests I felt were appropriate, but it's getting late and I'm having a knee surgery tomorrow so I'm sure the test coverage could be better. I wanted to put this out there for you to look over so we can hash out if this method is a good way forward or not.

Let me know what you think!

Resolves #12


mix.exs
```diff
@@ -26,7 +26,8 @@ defmodule JaSerializer.Mixfile do
      {:earmark, "~> 0.1", only: :dev},
      {:inch_ex, "~> 0.4", only: :docs},
      {:scrivener, "~> 1.0", optional: true},
-     {:ex_doc, "~> 0.7", only: :dev}]
+     {:ex_doc, "~> 0.7", only: :dev},
+     {:mix_test_watch, "~> 0.2", only: :dev}]
```
hunterboerner: @alanpeabody you sure you want this in JaSerializer's code?

green-arrow: Ah whoops, completely forgot I had that in there. Will gladly remove :)

alanpeabody:
Thanks a bunch @green-arrow I will take an in-depth look at this today, but in the meantime good luck with your surgery!

alanpeabody:
This is great! :+1: :100: :trophy:

I would probably like to remove the test reloading packages, although I may have to try them out.

green-arrow:
@alanpeabody thanks!

I went ahead and removed those deps. I'm used to having some kind of watch mechanism when I write code so everything auto-refreshes, instead of having to `mix test` every time. I would recommend giving them a try :)

alanpeabody:
Awesome, thanks!

PR #42, opened 2015-11-24
Updating README to reflect optional include / sparse fieldset

green-arrow:
Just let me know if there's anything you'd like to change


alanpeabody:
Looks great thanks!


As an aside, now that the base JSONAPI spec is covered I want to emphasize the protocol and figure out a way to make relationships work with it. I would like to begin to update the documentation to emphasize it as well.

PR #46, opened 2015-12-15
Refactor link builder to support more options on the link params parsing

simonprev:
I needed to support link like `"comments?article_id=:id"`. So I refactored the `path_for_context` function in the `Builder.Link` module.

The function now replace every occurence of `:` +_n_ word character (so it stops matching on `&` or `/`) and replace it with the `frag_for_context` return. As the regex always match with `:` the empty `frag_for_context` function is not needed anymore!

### DIFF
--- a/lib/ja_serializer/builder/link.ex
+++ b/lib/ja_serializer/builder/link.ex
@@ -1,6 +1,8 @@
 defmodule JaSerializer.Builder.Link do
   @moduledoc false
 
+  @param_fetcher_regex ~r/:\w+/
+
   defstruct href: nil, meta: nil, type: :related
 
   def build(_context, _type, nil), do: nil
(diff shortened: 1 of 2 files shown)


alanpeabody:
@simonprev great thanks, this makes more sense that what is there.

As an aside, I would like to also add better support for Phoenix router helpers, is that what you are using to generate URLs?

simonprev:
You mean calling the Phoenix router helper instead of specifying the `@location` using a string?

I’m open to work on better Phoenix support as it is the framework I use in all my projects :smile:

alanpeabody:
Yes, I was thinking as an alternative to the string, not replacing it. The challenge is supporting all the possible arguments to Pheonix routes.

It may just be easiest to accept a function that gets model and conn as arguments:

```elixir
# Query param
has_many :comments,
 link: &(comments_path(&2, :index, article_id: &1.id))

# Nested
has_many :comments,
  link: &(article_comments_path(&2, :index, &1.id))
```

PR #48, opened 2015-12-16
Skip content-type verification on DELETE requests

mikgra:
Some frontend frameworks, such as [Ember](http://emberjs.com), skips the content-type header when there is no data to be sent. The current ContentTypeNegotiation plug halts all such requests.

As far as I can tell this change complies with the [specification](http://jsonapi.org/format/#crud-deleting).


alanpeabody:
Thanks @mikgra!

PR #49, opened 2015-12-22
Add all required dependencies to applications list

dmarkow:
I'm using [exrm](https://github.com/bitwalker/exrm) to deploy my Phoenix app, and per their [common issues page](https://github.com/bitwalker/exrm/blob/master/docs/Common%20Issues.md), all dependencies must be listed in the `:applications` block. In addition, if a dependency's `mix.exs` file doesn't list _its_ dependencies, you need to list those as well. Otherwise, those dependencies' modules won't be available.

This is the case with `Inflex`. When using ja_serializer in production, I was having errors as soon as it tried to use the Inflex module. I fixed it by manually adding `:inflex` to the `:applications` list in my own app, but based on that common issues page, and looking at the `:applications` list of several other popular packages (postgrex, ecto, httpoison, etc.), ja_serializer should probably list its required dependencies under `:applications`.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -15,7 +15,7 @@ defmodule JaSerializer.Mixfile do
 
   # Configuration for the OTP application
   def application do
-    [applications: [:logger]]
+    [applications: [:logger, :inflex, :plug, :ecto, :poison]]
   end
 
   defp deps do


alanpeabody:
Uhg, totally one I have hit as well using exrm. Not sure why I didn't update this.

As an aside, I was thinking about dropping Inflex in particular and requiring the key be `data:` or `model:` or something. Thoughts?

dmarkow:
Phoenix's default `mix phoenix.gen.json` output automatically puts lines like `render(conn, "index.json", foos: foos)` in the controller so continuing to support that would mean one less thing to change.

On the other hand, the more I work with Elixir the more I appreciate the tendency toward being explicit and eliminating magic when possible. When I first started using this package, I noticed I could use any one of `:data`, `:model`, `:foos`, etc., without knowing exactly which one was "right" (the current documentation shows examples using `:model` but doesn't mention that you can also use the model name). As a result, I have some controllers/views using one approach and others using another. Having a single `:data` or `:model` key and not allowing anything else would make the options to `render` very explicit and would encourage consistency across the documentation, end-user code, tests, etc. (not to mention removing the dependency on Inflex).

alanpeabody:
Thanks for the feedback. I will work on consistency in the examples and docs.

Still waffling on the `foos:` version. :-/

PR #54, opened 2016-01-06
Resolve pipe warnings

bortevik:
Resolve warnings of piping into a function call without parentheses.


alanpeabody:
Thanks! (don't know why travis is having issues all of a sudden)

PR #60, opened 2016-01-12
Format serializer `type` based on config format

### LINKED ISSUE #27: Default type doesn't properly translate camel case words
If I have a `MyApp.PostView` module, the `type` in the output is correctly extracted from the module name and set to `post`. However, if I have `MyApp.BookStoreView`, the `type` is set to `bookstore` instead of `book_store` as I would expect.

`Mix.Utils` includes an approach at an `underscore` method, though it's lengthy at 11 function clauses. We could also do a very basic approach and just use `String.replace` with a Regex, prefixing any capital letters with an underscore.

In the mean time, adding a `def type, do: "book_store"` works.

linstula:
Closes #27

This PR formats at resource's type based on the format options specified
in `config.exs`.

Defaults to dasherized type.

Specifying:

```exs
config :ja_serializer,
  key_format: :underscored
```

would result in MyApp.BookStoreView returning a type of `book_store`

### DIFF
--- a/lib/ja_serializer/formatter/utils.ex
+++ b/lib/ja_serializer/formatter/utils.ex
@@ -29,11 +29,18 @@ defmodule JaSerializer.Formatter.Utils do
   def format_key(k) when is_atom(k), do: k |> Atom.to_string |> format_key
   def format_key(key), do: do_format_key(key, @key_formatter)
 
-
   @doc false
   def do_format_key(key, :underscored), do: key
   def do_format_key(key, :dasherized), do: String.replace(key, @dasherize, "-")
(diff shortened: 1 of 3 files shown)


linstula:
@alanpeabody let me know how you feel about this implementation and if there is anything you'd like to see changed.

alanpeabody:
@linstula this is great! It is faster the the existing regex logic I was using, is there any reason not to replace the `do_format_key/2` implementation as well?

alanpeabody:
This should work great for keys too:

https://github.com/AgilionApps/ja_serializer/commit/52225abdef1315e393470a449d3962385a446d40

linstula:
:tada: awesome!

PR #66, opened 2016-01-25
Format query params for pagination links based on config format

### LINKED ISSUE #64: Format query params
Currently, pagination links produce a link with underscored query params Eg:
`"/api/v1/posts?page[page]=1&page[page_size]=20"`

I wanted to get your thoughts on how you feel about formatting query params to be consistent with `key_format` specified in config.exs so that if you specified:

```exs
config :ja_serializer,
  key_format: :dasherized
```

the resulting link would be: `"/api/v1/posts?page[page]=1&page[page-size]=20"`.
Thoughts?

linstula:
Closes #64 

This PR formats the query params in pagination links to be consistent with the configured `key_format`. It also will format dasherized query params to underscores during deserialization.

Note: Because the query params were always underscored and because the default `key_format` is dasherized, this may cause issues for people that were relying on this bug (Eg. they expect underscored query params even though `key_format` is configured to `:dasherized`).

### DIFF
--- a/lib/ja_serializer/deserializer.ex
+++ b/lib/ja_serializer/deserializer.ex
@@ -35,7 +35,9 @@ if Code.ensure_loaded?(Plug) do
     @behaviour Plug
 
     def init(opts), do: opts
-    def call(conn, _opts), do: Map.put(conn, :params, format_keys(conn.params))
+    def call(conn, _opts) do
+      Map.merge(conn, %{params: format_keys(conn.params), query_params: format_query_params(conn.query_params)})
+    end
 
     defp format_keys(%{"data" => data} = params) do
       Map.merge(params, %{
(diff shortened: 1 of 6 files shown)


lib/ja_serializer/deserializer.ex
```diff
@@ -35,7 +35,11 @@ if Code.ensure_loaded?(Plug) do
     @behaviour Plug
 
     def init(opts), do: opts
-    def call(conn, _opts), do: Map.put(conn, :params, format_keys(conn.params))
+    def call(conn, _opts) do
+      conn
+      |> Map.put(:params, format_keys(conn.params))
```
bcardarella: Might be better to use `Map.merge/2` instead:

```elixir
conn
|> Map.merge(%{params: format_keys(conn.params), query_params: format_query_params(conn.query_params)})
```

linstula:
Found a bug in this branch - closing for now.

PR #67, opened 2016-01-26
Format query params

### LINKED ISSUE #64: Format query params

linstula:
Closes #64

This PR formats the query params in pagination links to be consistent with the configured `key_format`. It also will format dasherized params to underscores during deserialization.

Note: Because the pagination query params were always underscored and because the default key_format is dasherized, this may cause issues for people that are consuming pagination links that were relying on this behavior (Eg. they expect underscored query params even though key_format is configured to :dasherized).

### DIFF
--- a/lib/ja_serializer/builder/scrivener_links.ex
+++ b/lib/ja_serializer/builder/scrivener_links.ex
@@ -46,7 +46,7 @@ if Code.ensure_loaded?(Scrivener) do
 
     defp page_url({key, val}, %{opts: opts, conn: conn, model: page}) do
       base = opts[:page][:base_url] || conn.request_path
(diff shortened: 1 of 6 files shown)


alanpeabody:
Thanks @linstula. :star: 

I will put this out as 0.7.0 once I get a chance to update the README and changelog.

linstula:
Thanks! :confetti_ball: :tada:

PR #68, opened 2016-01-27
Improve deprecation messages

derekprior:
The deprecations logged are inconsistent and don't contain enough
information to quickly identify the offending code. This change makes
them consistent in how they are displayed and also includes stacktrace
information in the deprecation, which makes it simple to find the
source.

Previously some deprecations forced ANSI coloring while others had no
coloring at all. This change makes them all colored consistently but
only applies the coloring if elixir detects that it is supported.

To get a useful stacktrace out of `normalize_relation_opts`, I had to
pass the `__CALLER__` in from the macro to that module function so it
could be used in the deprecation. There was seemingly no way to get at
the information otherwise.

### DIFF
--- a/lib/ja_serializer/serializer.ex
+++ b/lib/ja_serializer/serializer.ex
@@ -56,7 +56,6 @@ defmodule JaSerializer.Serializer do
   """
   defcallback id(model, Plug.Conn.t) :: id
 
-
   @doc """
   The type to be used in the resource object.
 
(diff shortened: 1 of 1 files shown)


derekprior:
Here's an example from an app I'm working on:
terminal_ _tmux_ _tat_ _130x40

alanpeabody:
Awesome, thanks @derekprior !

Issue #69, opened 2016-01-30
Deserializing error


zipme:
In the latest release (0.7.0), when the request body contains `%Ecto.Date{}` or `%Plug.Upload{}`, an error occurred: 

```elixir
  data = %{
     attributes: %{
       attachment: %Plug.Upload{path: "image.png", filename: "image.png"}
    }
  }

  conn |> put(post_path(conn, :update, post), data: data)
```

```elixir
** (Protocol.UndefinedError) protocol Enumerable not implemented for %Plug.Upload{...}
```

The code above runs on okay with version 0.6.2.

alanpeabody:
Thanks, I will get a test and fix up asap

alanpeabody:
@zipme just looking into this. How would an Ecto.Date make it in via params?

zipme:
@alanpeabody thanks for looking into it! I think I was doing something wrong here, instead of passing Ecto.Date struct in the params in my test I should be using date string.

PR #70, opened 2016-01-30
Add ParamParser protocol to handle parsing params.

alanpeabody:
This allows anything parsed as a struct (eg from another plug or just because it has a __struct__ key passed along) to be implemented by the user.

@linstula @zipme.

I still am a bit unsure how this should really behave and looking for feedback, but I was thinking a protocol would be a super easy way to solve #69 and have an easy way to solve anything similar that comes up.

I am particularly unsure if nested maps in attributes should have their keys formatted. Thoughts?

I also haven't looked at any possible performance impact (anything passing through Any will be slow).


zipme:
❤️

Issue #63, opened 2016-01-19
key_format config not working


icn:
I have put the following lines to config.exs of phoenix framework 

config :ja_serializer,
  key_format: :underscored

But I still got dash-erized json response like this

{"jsonapi":{"version":"1.0"},"data":{"type":"maxbid","links":{"self":"/invocies/115236"},"id":"115236","attributes":{"user-id":2, "invoice-id": 115236}}}

 I would like to get 

{"jsonapi":{"version":"1.0"},"data":{"type":"maxbid","links":{"self":"/invocies/115236"},"id":"115236","attributes":{"user_id":2, "invoice_id": 115236}}}

This is  api pipeline

 pipeline :api do
    plug :accepts, ["json"]
    plug JaSerializer.Deserializer
  end 

What else I need to do to get underscored  json response ?

alanpeabody:
Thanks for your report @icn, I will look into it. Can you let me know if you run `mix deps.clean ja_serializer && mix deps.get` if the problem is resolved?

icn:
@alanpeabody  the issue is gone : ) after running `mix deps.clean ja_serializer && mix deps.get`

alanpeabody:
Great, that makes sense, I am storing the value in a module attribute at compile time. I will re-think if that is the best way.

PR #76, opened 2016-02-23
Update README to recompile after adding key_format

elijahkim:
taking information I found in https://github.com/AgilionApps/ja_serializer/issues/63 to clarify how to format keys.


alanpeabody:
Thanks!

PR #78, opened 2016-03-05
Fixes UndefinedFunctionError for parsing lists

beerlington:
This fixes an issue where the parser was trying to call format/1 instead
of parse/1 when parsing lists.


alanpeabody:
:tada: Thanks!

PR #89, opened 2016-03-28
Allow Multiple Links for each Relationship

benfalk:
## Overview

The JSON-API spec specifies that if a Relationship object includes
the `links` key it must have at least `self` or `related` as one of
it's link keys: http://jsonapi.org/format/#document-resource-object-relationships

The `link` option for the `has_one` and `has_many` relationships
under the serializer were forcing it to always be placed under the
`related` key.  This change allows you specify both the `related`
and `self` links - along with any abitrary key links you want to
have why still backwards supporting the `link` option.

## Oustanding Questions

Is this the right way to go about doing this?  If so I may need some
guidance on what I should change in README and internal comment
docs regarding this change.

### DIFF
--- a/lib/ja_serializer/builder/relationship.ex
+++ b/lib/ja_serializer/builder/relationship.ex
@@ -17,10 +17,12 @@ defmodule JaSerializer.Builder.Relationship do
   end
 
   defp add_links(relation, {_type, _name, opts}, context) do
-    case opts[:link] do
-      nil ->  relation
-      path -> Map.put(relation, :links, [Link.build(context, :related, path)])
-    end
+    Keyword.get(opts, :links, [])
+      |> Enum.map(fn {key, path} -> Link.build(context, key, path) end)
+      |> case do
+        []   ->  relation
+        links -> Map.put(relation, :links, links)
+      end
   end
 
   defp add_data(relation, {_t, name, opts}, context) do
--- a/test/ja_serializer/builder/relationship_test.exs
+++ b/test/ja_serializer/builder/relationship_test.exs
@@ -22,7 +22,13 @@ defmodule JaSerializer.Builder.RelationshipTest do
 
   defmodule FooSerializer do
     use JaSerializer
-    has_many :bars, type: "bar"
+    has_many :bars,
+             type: "bar",
+             links: [
+               self: "/foo/:id/relationships/bars",
+               related: "/foo/:id/bars"
+             ]
+
     has_one :baz, field: :baz_id, type: "baz"
     def bars(_,_), do: [1,2,3]
   end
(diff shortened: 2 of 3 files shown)


alanpeabody:
Thanks for the PR @benfalk.

I will dig into this a bit more this evening.

lib/ja_serializer/builder/relationship.ex
```diff
@@ -17,10 +17,14 @@ defmodule JaSerializer.Builder.Relationship do
   end
 
   defp add_links(relation, {_type, _name, opts}, context) do
-    case opts[:link] do
-      nil ->  relation
-      path -> Map.put(relation, :links, [Link.build(context, :related, path)])
-    end
+    Keyword.get(opts, :relationship_links, [])
+      |> Keyword.put_new(:related, opts[:link])
+      |> Enum.filter(fn {_key, value} -> is_binary(value) end)
```
alanpeabody: I don't believe we need this filter. Atoms are also allowed (to use a function to return the link). I also have been meaning to add support for anonymous functions.

test/ja_serializer/builder/relationship_test.exs
```diff
@@ -22,7 +22,13 @@ defmodule JaSerializer.Builder.RelationshipTest do
 
   defmodule FooSerializer do
     use JaSerializer
-    has_many :bars, type: "bar"
+    has_many :bars,
+             type: "bar",
+             relationship_links: [
```
alanpeabody: I kind of feel like the "relationship" is a bit of a given here, what do you think about just "links"?

benfalk: I originally had `links` and flip-flopped on what to name the key.  I am totally fine with just "links"

lib/ja_serializer/builder/relationship.ex
benfalk: I was doing that to weed out a nil that can creep in from putting in the `link` key on 21.  I've been giving this some though and I wonder if it would be better to move that logic up into the area where the other opts are normalized ?

alanpeabody: Yeah, actually moving the optimization to the serializer makes a lot of sense.

benfalk: I'll redo that and update this PR :+1:

benfalk:
I've updated the PR a bit, let me know what your thoughts are @alanpeabody when you get a chance.

Thanks!

alanpeabody:
Awesome, looks great! If you want to take a stab at just adding a note about it to the docs and readme please feel free, otherwise I will do so before I release the next version.

benfalk:
Sure thing, I'll give it a shot sometime later today :smile:

PR #90, opened 2016-03-29
Adds Phoenix Generater Mix Task

Dreamer009:


alanpeabody:
Thanks @Dreamer009 !!!

Honestly I have to think about this one a bit. I really like the idea of a generator, but I have to think about some of the stuff in the PhoenixJsonApiHelper. In particular there is some overlap with another library we have in dev/early testing, https://github.com/AgilionApps/ja_resource.

Dreamer009:
ja_resource looks really nice, and I can see the overlap. I am totally cool with using a different library instead of my PhoenixJsonApiHelper. Thanks for writing ja_serializer. It has helped out my team a lot.

alanpeabody:
Glad it is helping. I am not ready to give up on this yet, I just need some time to think and play with it before I know what direction I would like to go.

Dreamer009:
Sounds good.

alanpeabody:
@Dreamer009 upon further reflection I feel most comfortable with a generator that is as close to the base phoenix json generator as possible with out introducing anything new.

I would be happy to look at PRs for other features/helpers independently.

My reasoning is that there are a couple of other libraries with different approaches to the controller side of this which are used in conjunction with ja_serializer.

If you are willing I would love to see a "vanilla" version of what you have here, if not I will probably pull your branch and iterate on it myself. 

Thanks again for you time and contribution!!!

Dreamer009:
I could definitely help out. How "vanilla" are you talking?

PhoenixJsonApiHelper.where_params is easy enough to remove, but without something alone the lines of PhoenixJsonApiHelper.to_params the generated controller's create and update methods will not follow the JSON API spec. I would love for there to be a preferred approach on the controller side, so we could use that in the generator. 

Also worth noting is the fact that the user can overwrite the templates by placing new ones under the 'priv/templates/phoenix.gen.json_api/' folder in their local projects. So it should be relatively easy for users to augment the generated code regardless of what controller package they decide to use.

lib/ja_serializer/phoenix_json_api_helper.ex
```diff
@@ -0,0 +1,132 @@
...
+  defp keys_to_symbols(obj) when is_map(obj) do
+    for k <- Map.keys(obj), into: %{} do
+      v = Map.get(obj, k)
+
+      {key_to_atom(k), v}
+    end
+  end
+
+  defp key_to_atom(key) when is_atom(key),      do: key
+  defp key_to_atom(key) when is_bitstring(key), do: String.to_atom(key)
```
alanpeabody: Just getting back to this. I am planning on pulling this down into a branch and making some changes (mostly around name spaces/naming), but I wanted to let you know that this exposes a vulnerability anywhere it is used.

Every atom is stored indefinitely, up to a (default) max of 1048576. A user could maliciously send requests with lots of different keys and fill this up, breaking your app.

Dreamer009:
@alanpeabody very good note. I agree that using strings would be a better approach here. I am curious though, is that limit held by Elixir or a single Phoenix process?

Also I saw your merge request for JaSerializer.Params I can work on creating a branch that uses your helper methods instead. Maybe after I get off work today, or sometime later this week.

alanpeabody:
That limit is global, see http://erlang.org/doc/efficiency_guide/commoncaveats.html#id57371 for more info.

alanpeabody:
Great, the one thing I want to tweak is naming. I sort of feel like the generator should be name spaced to the library in some way. Thoughts?

Dreamer009:
I updated my branch to use your helper method.

I have mixed feelings on what the generator should be called. I like having it under the mix phoenix.gen list so people understand that it mirrors the phoenix template structure. At the same time I understand that it is misleading. If you give me a different name for the task I can update my code to use that name instead.

lib/mix/tasks/phoenix.build_templates.ex
```diff
@@ -0,0 +1,42 @@
+defmodule Mix.Tasks.Phoenix.BuildTemplates do
```
alanpeabody: What does this file do?

Dreamer009: It is an extra file I accidentally copied from phoenix. Give me a second, and I'll remove it.

alanpeabody:
@Dreamer009 I have been thinking about this and want to get it in for the 0.9.0 release.

I also am of mixed feeling on the name. However I do feel like it needs to include ja_serializer so we can make sure it doesn't conflict with any other library.

I was thinking something like:

```shell
mix ja_serializer.gen.phoenix_resource
mix ja_serializer.gen.phoenix_api
```

This is nice because it lets Phoenix maintain it's own namespace, while this generator becomes pretty explicit.

The only last issue to deal with then is will be wrapping any code (probably the whole modules) that depends on Phoenix in `Code.ensure_loaded?(Phoenix)` so this library can continue to compile on projects where phoenix is not installed.

Dreamer009:
@alanpeabody I made your suggested changes. I do have a few notes. I am not sure if they should be in separate issues.

1) We have been using your JaSerializer.Params code in our app and found a small bug. Line 60 should change from:

`({_name, %{"data" => nil}}, rel) -> rel`

to:

`({name, %{"data" => nil}}, rel) ->
  Map.put(rel, "#{name}_id", nil)
`
2) It was brought to my attention by a coworker that under http://jsonapi.org/format/#document-top-level that the json api allows for an array of resource objects. My current implementation does not allow for this... any thoughts?

alanpeabody:
1. would belong in it's own PR.
2. I don't quite follow, where is this an issue?

On the PR itself, it looks good, but I noticed a bit of naming inconsistency between file names and the module name.

Dreamer009:
I cleaned up that file name.

From the JSON API Org page:

>Primary data MUST be either:

1) A single resource object, a single resource identifier object, or null, for requests that target single resources
2) An array of resource objects, an array of resource identifier objects, or an empty array ([]), for requests that target resource collections

I handle the first case if we the user updates with a single object via:

`{
  "data": {
    "type": "articles",
    "id": "1",
    "attributes": {name: "thing 1"},
    "relationships": {}
  }
}`

but not the case where a user want to update multiple records at once. I have not ran into this on our project, but I believe the request would look something like:

`{
  "data": [{
      "type": "articles",
      "id": "1",
      "attributes": {name: "thing 1"},
      "relationships": {}
    },
    "type": "articles",
      "id": "2",
      "attributes": {name: "thing 2"},
      "relationships": {}
  }]
}`

alanpeabody:
Ah, yeah I didn't realize you meant when updating.

I think that ends up as a custom action. Lets not worry about it for now.

priv/templates/ja_serializer.gen.phoenix_api/controller.ex
```diff
@@ -0,0 +1,60 @@
+defmodule <%= module %>Controller do
+  use <%= base %>.Web, :controller
+
+  alias <%= module %>
+  alias JaSerializer.Params
+
+  plug :scrub_params, "meta" when action in [:create, :update]
```
beerlington: The "meta" attribute is optional according to the JSON API spec so this isn't necessary. I think you'll need to update the create and update functions to account for this.

priv/templates/ja_serializer.gen.phoenix_api/controller.ex
```diff
@@ -0,0 +1,60 @@
...
+  plug :scrub_params, "data" when action in [:create, :update]
```
beerlington: @alanpeabody Do you think this is necessary? With it, you get an error like `expected key "data" to be present in params, please send the expected key or adapt your scrub_params/2 call`, and without it, you'll get an error that there are no matching functions.

Maybe not necessary, but I think it provides a more friendly error message.

priv/templates/ja_serializer.gen.phoenix_api/controller.ex
```diff
@@ -0,0 +1,60 @@
...
+        |> render("show.json", data: <%= singular %>)
+      {:error, changeset} ->
+        conn
+        |> put_status(:unprocessable_entity)
+        |> render(<%= base %>.ChangesetView, "error.json", changeset: changeset)
+    end
+  end
+
+  def show(conn, %{"id" => id}) do
+    <%= singular %> = <%= alias %> |> Ecto.Query.where(id: ^id) |> Repo.one!
```
beerlington: What do you think of simplifying this to:

```ex
<%= singular %> = <%= module %>.Repo.get(<%= alias %>, id)
```

alanpeabody: :+1: For update and delete too.

priv/templates/ja_serializer.gen.phoenix_api/controller.ex
alanpeabody: I personally don't use them when building APIs. I could go either way here.

Dreamer009:
I addressed the comments left on the controller template.

alanpeabody:
Thanks @Dreamer009!

beerlington:
@Dreamer009 Can you squash your commits before we merge? Thanks!

Dreamer009:
@beerlington I squashed my commits

beerlington:
Looks good. Thanks @Dreamer009!

Dreamer009:
@alanpeabody @beerlington Thanks for all the help!

PR #94, opened 2016-04-15
Add Dynamic Type Functionality

benfalk:
An object is considered unique by combining it's `id` with its `type`
in the JSON API spec: http://jsonapi.org/format/#document-resource-objects

> Within a given API, each resource object's type and id pair MUST
> identify a single, unique resource.

When presenting multiple "liked" items that could collide with each
other via the `id` this becomes a valueable tool to have one serializer
that can present all similar items and provide a more approriate `type`
so that each item can be considered unique.


alanpeabody:
Awesome, thanks!

PR #95, opened 2016-04-15
Extend Dynamic Type Functionality to Relationships

benfalk:
When specifing a Serializer for a `has_one` or `has_many` relationship,
it will now provide the correct dynamic relationship calculated by
the given fun.

Sorry @alanpeabody, I jumped the gun a bit and didn't realize that this change wouldn't trickle down to the relationships.  This change will shore up that short-fall.


alanpeabody:
Great! I will push out a release this weekend with this.

PR #96, opened 2016-04-18
When Scrivener page has 0 total pages only the self link should be rendered

adamboas:
When the page object is passed into the scrivener builder it can have total_pages of 0 and page_number as 1. This is the case when there were no results returned by the query. In this case there should not be any next, last or previous links.


alanpeabody:
Awesome, thanks!

PR #98, opened 2016-04-24
Adds param utility for merging attrs and relations.

alanpeabody:
Does not handle whitelisting and keeps keys as strings to mitigate atom
overflow attacks.

@Dreamer009 this is a first step to allowing use of generators. I had initially thought about just generating "create_params" and "update_params" functions that manually generated the acceptable attributes, however I think you are correct and and a helper function to do this would be handy.

I ended up pulling in an implementation from JaResource because I wanted to address the atom issue as well as support multiple ids (many_to_many) and tweak the API to take the data json directly.

I am interested in feedback, in particular I am unsure if the name of the function really represents what it does, but `JaSerializer.Params.merge_attributes_and_relationships` seemed a little long.

### DIFF
--- a/lib/ja_serializer/params.ex
+++ b/lib/ja_serializer/params.ex
@@ -1,0 +1,71 @@
+defmodule JaSerializer.Params do
+  @moduledoc """
+  Functions to help when working with json api params.
+  """
+
+  @doc """
+  Takes the entire params passed in and merges relationships and attributes.
+
+  Note, this expects to recieve the json api "data" param.
+
+  Example functionality:
+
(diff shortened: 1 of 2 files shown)


beerlington:
Code LGTM. In terms of naming, flatten makes sense, but if you want to be pedantic, it's not just flattening or merging since it handles the *_id relationship key naming. You're essentially converting everything to a map of attributes so maybe some other possible names: `to_map`, `to_attributes`, `flatten_attributes`, or maybe even a generic `parse` or `parse_data`.

alanpeabody:
Hmm, I think I might go with to_attributes.

PR #100, opened 2016-04-27
add support for additional links on the resource [other than self]

dgvncsz0f:
objective
=========

Allow users to define additional links [besides the self link] on the
resource. For example:

```
  {"data": {"links": {"foo": "/foo",
                      "bar": "/bar"
                     }
           }
  }
```

Currently there is the `location` macro but that creates the `self`
entry only.

changes
=======

In this PR I've added a new macro `links` which works like location
but allows you to name the links. The `location` macro continues
to work and can even be used together. All tests continue to work 
without modification and
I've included a couple of additional tests on
`test/ja_serializer/json_api_spec/links_and_location_test.exs`
[I couldn't find a better place]. An example of serializer using the
new syntax:

```
  defmodule FooSerializer do
    links [ self: "/foo",
            permalink: "/foo?19700101000000+00:00"
          ]
  end
```

### DIFF
--- a/lib/ja_serializer/builder/resource_object.ex
+++ b/lib/ja_serializer/builder/resource_object.ex
@@ -22,7 +22,7 @@ defmodule JaSerializer.Builder.ResourceObject do
       data:          context.data,
       attributes:    Attribute.build(context),
       relationships: Relationship.build(context),
-      links:         [Link.build(context, :self, serializer.__location)],
+      links:         Enum.map(serializer.__links, fn {type, uri} -> Link.build(context, type, uri) end),
       meta:          serializer.meta(context.data, context.conn)
     }
   end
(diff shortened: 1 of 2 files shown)


alanpeabody:
Hey @dgvncsz0f thanks for the contribution! I think the functionality you have here is great and should be included.

What do you think about making this a function (eg links/2) that accepts the model and conn? Then we could add links using Phoenix helpers or based on user etc. The api could be mostly the same but look like:

```elixir
def links(_model, _conn) do
  [
    bars: "/foo/:id/bars"
  ]
end
```

As I move this library forward I am actually trying to make sure that all of the DSLs/macros are optional, as I know there are a lot of strong opinions on DSLs. I would like to try to add new features as behaviour callbacks with default functions first and they decide if they need to be enhanced with a macro based DSL as a second step.

alanpeabody:
A callback was added for this in #116. You can now define `links/2` callback instead of using the `location/1` macro.

Issue #110, opened 2016-05-29
[WIP] RFC: Relationship Definition


alanpeabody:
## Problem

The existing relationship DSL leaves something to be desired. In particular there are a handful of problems.

1. The relationships can't be customized quite enough for some uses cases. This leads to people using multiple serializers when they otherwise would only use one. See #103 and #45 as example requests.
2. The DSL is confusing and it's behavior is complicated. Some of this complexity stems from the various  ways JSON-API permits relationships to be represented and returned. Some of it is incidental complexity/confusion added while trying to improve other aspects of the DSL. #107, #71, #47 are examples of some of the issues being seen.
3. The link support could be better then atoms. In particular better support for Phoenix views would be good.

## Proposal

Relationships needs to be simplified while retaining full flexibility. I believe this can be accomplished in two parts.

1. Add an underlying behaviour callback for "relationships" that can be implemented. This function should accept the model and conn while returning a map of relationship structs. The default function will be populated by the DSL if it is used.
2) Create an easy to understand, explicit, well documented relationship definition that can be generated by DSL or in the struct directly.

## Scenarios

The final solution must handle all of:

* Linked relationships 
  * Using phoenix router helpers
  * Using simple strings
  * Both self and related links
* Resource Identifiers
* Resource Identifiers with included data
* Selective includes via the `include` parameter 
* Dynamic types
  * default to relationship name
  * leverage related serializer if provided
  * determine type based on object being serialized

#### Links

*Simple String*

While this is the default support linking mechanism in JaSerializer < 1.0, is it something that should continue to be supported? The "magic" of symbols calling functions as it works now is less then ideal.

```elixir
# DSL
has_one :post, link: "/v1/posts/:post_id" # not apparent how post_id is populated.

# Behaviour
def relationships(model, conn) do 
  %{
    post: %HasOne{related: "/v1/posts/:post_id"}
  }
end

# Relationship Struct
%JaSerializer.Relationship.HasOne{
  id: nil,
  type: nil,
  data: nil,
  related: "/v1/posts/:post_id",
  self: nil
}
```

*Anonymous function (Phoenix helpers)*
```elixir
# DSL
has_many :comments, link: &post_comment_url(&2, &1.id)

# Behaviour
def relationships(model, conn) do 
  %{
    comments: %HasMany{related: &post_comment_url(&2, &1.id)}
  }
end

# Relationship Struct
%JaSerializer.Relationship.HasMany{
  each_id: :id,
[...]

PR #116, opened 2016-06-05
First pass at making DSL completely optional.

alanpeabody:
* Seperates DSL logic into own module.
* Adds links/2 callback.
* Adds relationships/2 callback.
* Adds HasMany and HasOne structs.

Starts the behaviour part of the work specified in #110. Further enhancements and DSL tweaks to come in subsequent PRs.

### DIFF
--- a/lib/ja_serializer.ex
+++ b/lib/ja_serializer.ex
@@ -1,13 +1,30 @@
 defmodule JaSerializer do
   @moduledoc """
-  Provides a DSL for defining and how to serialize data to return in jsonapi.org 1.0 format.
+  Provides a behaviour and optional DSL for defining and how to serialize data.
 
-  See JaSerializer.Serializer for use.
+  The DSL can be opted out of by passing `dsl: false`, eg:
+
+      defmodule MyApp.PostView do
+        use JaSerializer, dsl: false
+      end
+
+  See JaSerializer.Serializer for details on the behaviour.
+
+  See JaSerializer.DSL for details on the DSL.
   """
 
-  defmacro __using__(_) do
-    quote do
-      use JaSerializer.Serializer
+  defmacro __using__(opts \\ []) do
+    # Default to using DSL for now.
+    opts = Keyword.put_new(opts, :dsl, true)
+    if opts[:dsl] do
+      quote  do
+        use JaSerializer.Serializer
+        use JaSerializer.DSL
+      end
+    else
+      quote do
+        use JaSerializer.Serializer
+      end
     end
   end
 end
--- a/lib/ja_serializer.ex
+++ b/lib/ja_serializer.ex
@@ -33,7 +33,7 @@ defmodule JaSerializer do
   """
 
   @doc false
-  defmacro __using__(opts \\ []) do
+  defmacro __using__(opts) do
     # Default to using DSL for now.
     opts = Keyword.put_new(opts, :dsl, true)
     if opts[:dsl] do
--- a/lib/ja_serializer.ex
+++ b/lib/ja_serializer.ex
@@ -1,6 +1,6 @@
 defmodule JaSerializer do
   @moduledoc """
-  A library for generating JSON-API.org spec complaint JSON.
+  A library for generating JSON-API.org spec compliant JSON.
 
   JaSerializer has three main public components:
 
--- a/lib/ja_serializer.ex
+++ b/lib/ja_serializer.ex
@@ -1,18 +1,38 @@
 defmodule JaSerializer do
   @moduledoc """
-  Provides a behaviour and optional DSL for defining and how to serialize data.
+  A library for generating JSON-API.org spec complaint JSON.
 
+  JaSerializer has three main public components:
+
+    * `JaSerializer.Serializer` - A Behaviour defining how a particular object
+      is serialized. Provides default empty callbacks and ability to override.
+
+    * `JaSerializer.DSL` - A macro based DSL that sits on top of the Behaviour
+      for quick, clean, and easy definition of serializers.
+
+    * `JaSerializer.PhoenixView` - Provides render functions for a view to take
+      advantage of the serialization interface.
+
+  When used the `JaSerializer` module includes the Behaviour and DSL by default.
   The DSL can be opted out of by passing `dsl: false`, eg:
 
-      defmodule MyApp.PostView do
+      defmodule MyApp.PostSerializer do
         use JaSerializer, dsl: false
+        # ...
       end
 
-  See JaSerializer.Serializer for details on the behaviour.
+  If using JaSerializer with Phoenix, your normal entry point is
+  `JaSerializer.PhoenixView`, eg:
 
-  See JaSerializer.DSL for details on the DSL.
+      defmodule MyApp.PostView do
+        use MyApp.Web, :view
+        use JaSerializer.PhoenixView
+        # ...
+      end
+
   """
 
+  @doc false
   defmacro __using__(opts \\ []) do
     # Default to using DSL for now.
     opts = Keyword.put_new(opts, :dsl, true)
--- a/lib/ja_serializer/dsl.ex
+++ b/lib/ja_serializer/dsl.ex
@@ -88,15 +88,15 @@ defmodule JaSerializer.DSL do
 
   defp define_default_relationships do
     quote do
-      def relationships(struct, conn) do
-        JaSerializer.DSL.default_relationships(__MODULE__, struct, conn)
+      def relationships(struct, _conn) do
+        JaSerializer.DSL.default_relationships(__MODULE__)
       end
       defoverridable [relationships: 2]
     end
   end
 
   @doc false
-  def default_relationships(serializer, struct, conn) do
+  def default_relationships(serializer) do
     serializer.__relations
     |> Enum.map(&dsl_to_struct/1)
     |> Enum.into(%{})
@@ -110,14 +110,14 @@ defmodule JaSerializer.DSL do
   defp define_default_links do
     quote do
       def links(data, conn) do
-        JaSerializer.DSL.default_links(__MODULE__, data, conn)
+        JaSerializer.DSL.default_links(__MODULE__)
       end
       defoverridable [links: 2]
     end
   end
 
   @doc false
-  def default_links(serializer, data, conn) do
+  def default_links(serializer) do
     %{self: serializer.__location}
   end
 
--- a/lib/ja_serializer/dsl.ex
+++ b/lib/ja_serializer/dsl.ex
@@ -1,6 +1,6 @@
 defmodule JaSerializer.DSL do
   @moduledoc """
-  A DSL for defining JSON-API.org spec complaint payloads.
+  A DSL for defining JSON-API.org spec compliant payloads.
 
   Built on top of the `JaSerializer.Serializer` behaviour.
 
@@ -14,16 +14,10 @@ defmodule JaSerializer.DSL do
   This module should always be used in conjunction with
   `JaSerializer.Serializer`, see `JaSerializer` for the best way to do so.
 
-  When `use`ing this module default implementations of the `links/2`,
-  `attributes/2`, and `relationships/2` will be defined on your module.
-
-  Overriding these callbacks can be a great way to customize your serializer
-  beyond what the DSL provides.
-
-  ## Example
+  ## DSL Usage Example
 
       defmodule PostSerializer do
-        use JaSerializer
+        use JaSerializer, dsl: true
 
         location "/posts/:id"
         attributes [:title, :body, :excerpt, :tags]
@@ -47,6 +41,11 @@ defmodule JaSerializer.DSL do
       |> PostSerializer.format
       |> Poison.encode!
 
+  When `use`ing JaSerializer.DSL the default implementations of the `links/2`,
+  `attributes/2`, and `relationships/2` callbacks will be defined on your module.
+
+  Overriding these callbacks can be a great way to customize your serializer
+  beyond what the DSL provides. See `JaSerializer.Serializer` for examples.
   """
 
   alias JaSerializer.Relationship.HasMany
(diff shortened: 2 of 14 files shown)


lib/ja_serializer.ex
```diff
@@ -1,13 +1,30 @@
...
+    opts = Keyword.put_new(opts, :dsl, true)
+    if opts[:dsl] do
+      quote  do
+        use JaSerializer.Serializer
+        use JaSerializer.DSL
+      end
+    else
+      quote do
+        use JaSerializer.Serializer
+      end
```
benfalk: looks like `JaSerializer.Serializer` is being used in both blocks.  What are your thoughts on move the `use` block outside of the if statement and only have a `use` statement for the dsl if it's requested?

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,287 @@
+defmodule JaSerializer.DSL do
+  alias JaSerializer.Relationship.{HasMany,HasOne}
```
benfalk: This is cool, what is this tuple like syntax and what version support does it have with elixir?

alanpeabody: Ha, it is 1.2 only, I just removed it to maintain backwards compatibility. Same as doing two aliases.

lib/ja_serializer.ex
alanpeabody: @benfalk I actually tried to clean this up a bit but didn't have too much luck. In fact this is what I ended up after spending way too much time trying to conditionally use inside the quote. What were you thinking exactly so I can give it a try?

alanpeabody:
This needs major documentation updates before merge.

lib/ja_serializer.ex
benfalk: Not sure if it works but this is what I was thinking
``` elixir
defmacro __using__(opts \\ []) do
  quote bind_quoted: [opts: opts] do
    opts = Keyword.put_new(opts, :dsl, true)
    use JaSerializer.Serializer
    if opts[:dsl], do: use JaSerializer.DSL
  end
end
```

alanpeabody: Unfortunately that doesn't work, the DSL is unavailable where it should be. I don't fully understand WHY it doesn't work, but that was pretty much my first approach too.

benfalk: that's a bummer, I've been running into some strange oddities as well with macros like that

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
+defmodule JaSerializer.DSL do
+  @moduledoc """
+  A DSL for defining JSON-API.org spec complaint payloads.
```
beerlington: *compliant

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+
+    * `location/1` - Define the url of a single serialized object.
+    * `attributes/1` - Define the attributes to be returned.
+    * `has_many/2` - Define a has_many relationship.
+    * `has_one/2` - Define a has_one or belongs_to relationship.
+
+  This module should always be used in conjunction with
+  `JaSerializer.Serializer`, see `JaSerializer` for the best way to do so.
+
+  When `use`ing this module default implementations of the `links/2`,
```
beerlington: module's ?

beerlington: nm, it's just missing a comma after module.

beerlington: Is it implied that these are "callbacks"? I got to the next sentence and wasn't clear whether it was talking about links, attributes and relationships or something else. Would it make sense to say that these are callbacks?

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+  `attributes/2`, and `relationships/2` will be defined on your module.
+
+  Overriding these callbacks can be a great way to customize your serializer
+  beyond what the DSL provides.
+
+  ## Example
```
beerlington: The previous paragraph talks about overriding callbacks, but this example does not do that. I was expecting the example to demonstrate that. Would it make sense to include that in the example?

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+
+      defmodule PostSerializer do
+        use JaSerializer
```
beerlington: Do you need `use JaSerializer.DSL` or is that automatically included?

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+
+        location "/posts/:id"
+      end
+
+      defmodule CommentSerializer do
+        use JaSerializer
+
+        location "http://api.example.com/posts/:post_id/comments/:id"
+
+        def post_id(comment, _conn), do: comment.post_id
```
beerlington: Is there a way to make this the default behavior? In this case it seems like you could just call comment.post_id if this function wasn't defined.

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+      end
+
+  ## Atom Example
+
+  When an atom is passed in it is assumed it is a function that will return
```
beerlington: To be consistent with the string version, I would maybe change this to:

```
When an atom is passed in, it is called as a function on the serializer with the struct and conn passed in.
```

lib/ja_serializer.ex
```diff
@@ -1,13 +1,50 @@
 defmodule JaSerializer do
   @moduledoc """
-  Provides a DSL for defining and how to serialize data to return in jsonapi.org 1.0 format.
+  A library for generating JSON-API.org spec complaint JSON.
```
bmfay: complaint --> compliant

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+      end
+
+  Links can be defined with an atom or string.
+
+  String may be either a relative or absolute path. Path segments beginning
+  with a colon are called as functions on the serializer with the struct and
+  conn passed in. In the above example id/2 would be called which is defined as
+  a default callback.
+
+  When an atom is passed in it is assumed it is a function that will return
```
beerlington: If you decide to change the location atom doc above, you'd want to update this too.

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+  details on the spec.
+
+  ## Link based relationships
+
+  Specify a URI or path which responds with the related resource. For example:
+
+      defmodule MyApp.PostView do
+        use JaSerializer
+
+        has_many :comments, link: :comments_link
```
beerlington: Do you think it's worth mentioning that this produces a "related" link?

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+      PostSerializer.format(post, conn, fields: field_param)
+
+      # via PhoenixView integrations from controller
+      render(conn, :show, data: post, opts: [fields: field_param])
+
+  ## Further customization
+
+  Further cusomization of the attributes returned can be handled by overriding
+  the `attributes/2` callback. This can be done in conjuction with the DSL
+  using super, or without the DSL just returning a map.
```
bmfay: cusomization -> customization, conjuction -> conjunction

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+    end
+  end
+
+  @doc """
+  Add a has_many relationship to be serialized.
+
+  JSONAPI.org supports three types or relationships:
+
+    * As links - Great for clients lazy loading relationships with lots of data.
+    * As "Resource Indetifiers" - A type/id pair, usefull to relate to data the client already has.
```
bmfay: indetifiers --> identifiers
usefull -> useful

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+        use JaSerializer
+
+        has_many :comments, serializer: MyApp.CommentView, include: false
+        has_many :tags, type: "tags"
+        has_one  :author, type: "user", field: :created_by_id
+
+        # ...
+      end
+
+  When you use the `has_many` and `has_one` macros an overridable function that
```
bmfay: One grammatical thing I might point out in a few places that is particularly relevant in technical documentation (since documentation is inherently hard to read due to how technical it is) would be to use commas at the end of clauses.  I would put a comma after macros here ^.

When you use the `has_many` and `has_one` macros, an overridable function that is expected to return the data for you is automatically defined.

lib/ja_serializer/dsl.ex
Dangeranger: Good question. The way I read the top level module documentation at https://github.com/AgilionApps/ja_serializer/pull/116/files#diff-bbd3e555da04d81d12cefe59568a05c5R20 was that unles `dsl: false` is included in the initial `use` statement that `JaSerializer.DSL` would be included automatically.

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+  It the tags example where just the `type` option is used the `id` field is
+  automatically used on each map/struct returned by the data source.
+
+  It is important to note that when accessing the relationship fields it is
+  expected that the relationship is preloaded. For this reason you may want to
+  consider using links for has_many relationships where possible.
+
+  ## Including related data
+
+  Returns a "Resource Indefifier" (see above) as well as the fully serialized
```
bmfay: *Identifier

lib/ja_serializer/dsl.ex
alanpeabody: going to make this explicitly `dsl: true`

lib/ja_serializer/dsl.ex
alanpeabody: Maybe? It is worth looking into anyway.

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+  relationship. The `field` option can be used to grab the id or models from a
+  different field in the serialized object. The author is an example of
+  customizing this, and is frequently used when returning resource identifiers
+  for has_one relationships when you have the foreign key in the serialized
+  struct.
+
+  In the comments example when a `serializer` plus `include: false` options are
+  used the `id/2` and `type/0` functions are called on the defined serializer.
+
+  It the tags example where just the `type` option is used the `id` field is
```
beerlington: It > In

lib/ja_serializer/dsl.ex
beerlington: Identifier

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+        use JaSerializer
+
+        has_many :comments, serializer: MyApp.CommentView, include: true
+        has_many :tags,     serializer: MyApp.TagView,     include: true
+        has_many :author,   serializer: MyApp.AuthorView,  include: true, field: :created_by
+
+        # ...
+      end
+
+  Just like when working with only Resource Indetifiers this will define a
```
beerlington: Identifiers

lib/ja_serializer/serializer.ex
```diff
@@ -1,47 +1,47 @@
 defmodule JaSerializer.Serializer do
   @moduledoc """
-  Define a serialization schema.
+  A Behaviour for defining JSON-API.org spec complaint payloads.
```
beerlington: super nitpick, but you're not consistent in the docs with how you refer to the JSON API spec. In some places it's "JSONAPI.org", others it's "JSONAPI spec".

lib/ja_serializer/serializer.ex
```diff
@@ -1,47 +1,47 @@
...
 
-  Provides `has_many/2`, `has_one/2`, `attributes/1` and `location/1` macros
-  to define how your data (struct or map) will be rendered in the
-  JSONAPI.org 1.0 format.
+  The following callbacks are available:
 
-  Defines `format/1`, `format/2` and `format/3` used to convert data for
-  encoding in your JSON library of choice.
+    * `id/2` - Return ID of struct to be specialized.
```
beerlington: Assuming you mean "serialized"

lib/ja_serializer/serializer.ex
```diff
@@ -1,47 +1,47 @@
...
+    * `type/0` - Return string type of struct to be specialized
```
beerlington: same here "serialized"

lib/ja_serializer/serializer.ex
```diff
@@ -1,47 +1,47 @@
...
-  ## Example
+  A Serializer (or view) is typically one of the few places in an API where
+  content and context are both present. To accomodate this each callback gets
+  the data being serialized (typically a struct, often called a model) and the
+  Plug.Conn as arguments. Context data such as the current user, role, etc
+  should typically be made available on the conn.
 
-      defmodule PostSerializer do
-        use JaSerializer
+  When `use`ing this module all callbacks get a default, overrideable
```
beerlington: *overridable

lib/ja_serializer/serializer.ex
```diff
@@ -77,51 +77,54 @@ defmodule JaSerializer.Serializer do
 
       def type, do: fn(model, _conn) -> model.type end
   """
+  # TODO: Can we convert this to type/2 for consistency without too much hassle?
   defcallback type() :: String.t | fun()
 
   @doc """
-  Returns a map of attributes to be mapped.
+  Returns a map of attributes to be included.
```
beerlington: I wonder if "included" should be "serialized" so it's not confused with the other notion of "included" in the spec.

lib/ja_serializer/dsl.ex
```diff
@@ -0,0 +1,424 @@
...
+  ## String Examples
+
+  String may be either a relative or absolute path. Path segments beginning
+  with a colon are called as functions on the serializer with the struct and
+  conn passed in.
+
+      defmodule PostSerializer do
+        use JaSerializer
+
+        location "/posts/:id"
```
Dangeranger: Both this example and the Atom example below feel like they should be overridable defaults. Being explicit has value too, but maybe just having these as concrete examples of how to override is enough for users to know if they need to customize.

lib/ja_serializer/dsl.ex
alanpeabody: I was really close on this one. 👀

PR #119, opened 2016-06-16
Fixed documentation that referenced non-existent `flatten`

jerel:
I can't find a function named `flatten` in the codebase so I'm guessing that the docs are what are wrong.


alanpeabody:
Whoops, thanks! Renamed this but missed it in the docs.

PR #120, opened 2016-06-17
Change callback type/0 to type/2.

alanpeabody:
type/0 will continue to work if defined, however type/2 now perfered.
type/0 returning an anonymous function with arity of 2 is deprecated.

This is more consistent and all callbacks are now an arity of two and
accept the data structure and the conn.


alanpeabody:
@benfalk this is a replacement for the work you introduced in #94. One of the last few changes I want to get in 0.10.0 before starting a 1.0 branch.

Issue #103, opened 2016-05-03
Optional relations


chrisjowen:
Hey There, 

Am I missing something but I don't see a way of optionally specifying a relationship. The include flag seems to specify whether to simply linking to relationship or to side load it, but I think it seems valid to me  in some cases to also do neither. Currently if I specify i.e. a has_one relationship in the serializer and I have not fetched the relationship then (assuming Ecto) this bombs when trying to render. 

It would be nice to be able to specify the relationship as optional and if the data has not ben fetched then the serializer will simply ignore when rendering.

alanpeabody:
Thanks for the feedback @chrisjowen. There have been a couple of requests oriented around this.

I will continue thinking about how to best support this kind of use case without making thing more complicated.

alanpeabody:
0.10.0 has support for this via the :identifiers option. More sophisticated situations can be handled with the relationships callback.

Issue #45, opened 2015-12-07
Selectively link or include a relation.


archSeer:
Hi, I've been having fun with this library in combination with Ember. :beers: for having it seamlessly "just work" out of the box!

I'd like to be able to selectively toggle between including or just linking to a relation. For example, in my app, I have an Album, that belongs to an Artist, and has many Tracks. I'd like to just link to the Track relation on `GET /albums`, and I'd like to sideload it on `GET /albums/:id`.

I've gotten close by using this:
```elixir
defmodule Colibri.AlbumView do
  use Colibri.Web, :view

  location "/albums/:id"
  attributes [:title]

  has_one :artist,
    serializer: Colibri.ArtistView,
    include: true

  has_many :tracks,
    serializer: Colibri.TrackView,
    link: "/albums/:id/tracks"
end

```

and then modifying the render call in the index route:

```elixir
render(conn, :index, albums: albums, opts: [include: "tracks"])
```

However this is not quite what I want, because it'll both link to the relationship and include the track ids in the `/album/:id` show action response (which makes Ember fetch these one-by-one, instead of just using the relation link).

I also tried dropping the `serializer` option, and manually doing `include` in the show action render call, however that breaks the code, since there's no serializer available.

I propose the following -- `include: false` (or maybe `embed: false`) would disable embedding:

```elixir
# serializer.ex
  has_many :tracks,
    link: "/albums/:id/tracks",
    serializer: Colibri.TrackView,
    include: false
```

And then, we'd be able to explicitly opt-in on the render action:
```elixir
# controller.ex
render(conn, :index, albums: albums, opts: [include: "tracks"])
# this could potentially include the default relations + tracks
render(conn, :index, albums: albums, opts: [include: "default,tracks"])
```

I've looked into the internals to try and write a patch for this, but it's my first week of using Elixir, so I'm not quite there yet...

alanpeabody:
Thanks for the feedback and issue report @archSeer.

Include should default to false, I would think what you have defined should work.

Then in your show action:

```elixir
render(conn, :index, albums: albums, opts: [include: "artist,tracks"])
```

that SHOULD result in a document something like (please excuse my poorly formatted json):

```javascript

{
  data: {
    id: 1,
    type: 'album',
    attributes: { /* ommited for breviety */ },
    relationships: {
      tracks: {
        related: '/albums/1/tracks',
[...]

PR #121, opened 2016-06-20
Allow un-included includable relations w/o data

alanpeabody:
Adds `identifiers` to the relationship options that controls if
identifiers are included when a relationship is not included.

This would address: #45 & #103


beerlington:
LGTM. I'd like to review the docs a little bit at a later time to make sure I understand it all, but the implementation seems 👍

PR #122, opened 2016-06-21
Deprecate MyMod.format/3 for JaSerializer.format/4

alanpeabody:
Previously we were defining a function on the module that was not a
callback default. We should not do that, so lets deprecate for removal
in 1.0.

Instead we use `JaSerializer.format(MyMod, data, conn, opts)`

### DIFF
--- a/lib/ja_serializer.ex
+++ b/lib/ja_serializer.ex
@@ -47,4 +47,16 @@ defmodule JaSerializer do
       end
     end
   end
+
+  @doc """
+  Main serialization method.
+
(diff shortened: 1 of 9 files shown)


beerlington:
:shipit:

hunterboerner:
Unfortunately this means that I can no longer pipe data into format. :(

alanpeabody:
Sorry... As a work around you can define format on your serializers yourself, but defining a function in using on a module that is not a default implementation of a call back is an elixir anti pattern.

PR #123, opened 2016-06-23
Fixes imperative assignment.

linjunpop:


alanpeabody:
Thank you, but I don't think this is quite right as I believe it could result in nil opts. Also the tests seem to be failing.

Maybe a case statement would be better here?

linjunpop:
You're right, I'll fix this.

parndt:
I believe #124 fixes this

alanpeabody:
Thanks, but #124 got to it first!

linjunpop:
Cool

PR #124, opened 2016-06-23
Re-wrote the if statement in DSL to be elixir 1.3 complient

itsgreggreg:
Otherwise you get a compile warning on imperative assignment inside of if blocks. :)


alanpeabody:
Thanks!

PR #125, opened 2016-06-24
Fixed warning in generator.

parndt:
Fixes this warning:

```
warning: the variable instructions is unsafe as it has been set inside a case/cond/receive/if/&&/||. Please explicitly return the variable value instead. For example:

    case int do
      1 -> atom = :one
      2 -> atom = :two
    end

should be written as

    atom =
      case int do
        1 -> :one
        2 -> :two
      end

Unsafe variable found at:
  lib/mix/tasks/ja_serializer.gen.phoenix_api.ex:90
```

Let me know if anything needs changing 👍


parndt:
I also did the same fix as #124 but that PR was first 😄

alanpeabody:
Thanks!

PR #127, opened 2016-06-27
per Slack conversation: guard against the use of json vs json-api 

nazarsh:
88 tests, 0 failures

Thanks again for your help in tracking down my resp_header issue!

### DIFF
--- a/lib/ja_serializer/phoenix_view.ex
+++ b/lib/ja_serializer/phoenix_view.ex
@@ -47,24 +47,12 @@ defmodule JaSerializer.PhoenixView do
     quote do
       use JaSerializer, unquote(opts)
 
-      def render("index.json", data) do
-        JaSerializer.PhoenixView.render(__MODULE__, data)
-      end
-
       def render("index.json-api", data) do
-        JaSerializer.PhoenixView.render(__MODULE__, data)
-      end
-
-      def render("show.json", data) do
         JaSerializer.PhoenixView.render(__MODULE__, data)
       end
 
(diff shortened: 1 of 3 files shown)


alanpeabody:
Going to hold off on merging this for just a bit to see if any 0.10.0 bugs come in.

This will go out as 0.11.0 because it might break things for people.

alanpeabody:
Actually, maybe it would be better to soft deprecate these with a warning instead of just removing them. We also need to update the generators to be correct, as they stand now they are wrong.

Sorry, I know I said removing was what should be done, but I think that will break things for a lot of people.

Thank you for bringing this to my attention, if you want to update this to be a soft deprecate (warning) and update the generators I will happily merge it. If will get something in over the next few days.

nazarsh:
This definitely makes sense - I will see if I can updates done tonight. Thanks for looking into this.

nazarsh:
I hope I didn't miss anything. I've duplicated the tests a bit just to make sure both cases are covered and still return the warnings. Sorry things are not as DRY as they could have been... Happy to make any additional changes though.

alanpeabody:
Looks great, thanks!

alanpeabody:
@nazarsh can you remember why we needed to do this?

nazarsh:
@alanpeabody I think I've triggered an edge case which resulted in incorrect response header (and we talked about not forcing a particular response since people might be wanting to override it)

alanpeabody:
Thanks!

PR #129, opened 2016-06-28
Substitute all values in Changeset errors

### LINKED ISSUE #128: JaSerializer.EctoErrorSerializer.format(changeset) leaves format variables in Ecto 2.0
Ecto 2.0 I think changed the format of Changeset error messages: The output from `JaSerializer.EctoErrorSerializer.format(changeset)` does not contain the human readable string, but instead the template string:

```
code: %Alembic.Error{detail: "Miles willing to travel must be less than or equal to 300", source: %Alembic.Source{pointer: "/data/attributes/miles-willing-to-travel"}, title: "must be less than or equal to 300"} in errors
lhs:  %Alembic.Error{code: nil, detail: "Miles willing to travel must be less than or equal to 300", id: nil, links: nil, meta: nil, source: %Alembic.Source{parameter: nil, pointer: "/data/attributes/miles-willing-to-travel"}, status: nil, title: "must be less than or equal to 300"}
rhs:  [%Alembic.Error{code: nil, detail: "Miles willing to travel must be less than or equal to %{number}", id: nil, links: nil, meta: nil, source: %Alembic.Source{parameter: nil, pointer: "/data/attributes/miles-willing-to-travel"}, status: nil, title: "must be less than or equal to %{number}"}]
```

`Alembic.Error` is from my library, but the contents of struct are decoded from the output of `JaSerializer.EctoErrorSerializer.EctoErrorSerializer.format(changeset)`, so I think `format` needs to behave differently for Ecto 1.X vs Ecto 2.X.

KronicDeth:
Fixes #128

Instead of special casing `%{count}`, use the recommended substitution from the [@doc for `Ecto.Changeset.traverse_errors`](https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/changeset.ex#L1836-L1838).  It produces the same format messages in Ecto 1.0 and Ecto 2.0 this way.

### DIFF
--- a/lib/ja_serializer/ecto_error_serializer.ex
+++ b/lib/ja_serializer/ecto_error_serializer.ex
@@ -11,11 +11,15 @@ defmodule JaSerializer.EctoErrorSerializer do
   end
 
   defp format_each({field, {message, vals}}) do
-    message = Regex.replace(~r/%{count}/, message, "#{vals[:count]}")
+    # See https://github.com/elixir-ecto/ecto/blob/34a1012dd1f6d218c0183deb512b6c084afe3b6f/lib/ecto/changeset.ex#L1836-L1838
+    title = Enum.reduce vals, message, fn {key, value}, acc ->
+      String.replace(acc, "%{#{key}}", to_string(value))
+    end
+
     %{
       source: %{ pointer: pointer_for(field) },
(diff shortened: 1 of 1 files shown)


alanpeabody:
Awesome, thanks!

PR #135, opened 2016-07-09
Update Phoenix controller generator to use recommended error rendering

kitlangton:
Rendering with the ChangesetView did not display the errors in the correct JsonApi format. Now it does!

(thanks for the help, Alan!)


alanpeabody:
Thanks!

Issue #130, opened 2016-06-28
If a model contains an attribute `type` it gets used as the resource type.


jerel:
I just updated to latest on master from 4d949cd21 and my resource type behavior is no longer correct. I have `attributes` defined like this: `attributes = [:name, :alias, :type]` in a PhoenixView and ja_serializer is using the value from the `type` attribute as the resource type. If I define a `type/2` introduced in d1a453dc59ed21bf8461fc0f8e3eb9b6ab75c685 then its returned value gets used for the resource but it is also used for the attribute value.

alanpeabody:
Oh, yeah, you are absolutely right.

I am honestly not really sure what to do about this, as I think changing type/0 to type/2 was important for consistency.

If you only have this in a few places I might suggest just not using the attributes/1 macro and instead use the attributes/2 callback:

```elixir
def attributes(struct, _conn), do: Map.take(struct, [:name, :alias, :type])
```

jerel:
Thanks, this is a nice enough solution. Would it be good to call out the naming conflict with the solution in the documentation? Happy to add it if you want it done.

Also, on the topic of documentation: although probably simple it's non-obvious to me how one would render with a different serializer in Phoenix. Given this example: `render(conn, :show, data: article)` what if I want to do something like `render(conn, ArticleWithoutRelationships, data: article)`? Is that possible?

alanpeabody:
Sure, noting that none of the callbacks can be used as attributes would be good. We could probably enforce that at compile time in the macro too.

alanpeabody:
To follow up on your second question, if you have an alternative view you can use https://hexdocs.pm/phoenix/Phoenix.Controller.html#render/4 to render to that view instead.

If your goal is to just not include (sideload) relationships you can look at using the serializer, link, and identifiers: :when_included option. An example might look like this:

```elixir
has_many :comments,
  serializer: CommentView,
  identifiers: :when_included,
  link: "/comments?filter[post]=:id"
```

to include it then just pass the include option:

```elixir
render(conn, data: article, opts: [include: "comments"])
```

deadtrickster:
is there any way to define render has_many relationship without preloading if all I want is just related link?

alanpeabody:
@deadtrickster yes, see: https://hexdocs.pm/ja_serializer/JaSerializer.DSL.html#has_many/2

The "Link based relationships" section has what you are looking for.

deadtrickster:
[...]

PR #137, opened 2016-07-10
Change the inclusion behavior when include is nil

alanpeabody:
When include is nil and the relationship is not in the include param the
identifiers are not included unless `indentifiers: :always` option is
set.

Fixes the second half of #130


beerlington:
Lgtm

PR #139, opened 2016-07-16
Use build_conn instead of conn in test blueprint

dustinfarris:
Deprecated in https://github.com/phoenixframework/phoenix/issues/1486


alanpeabody:
Thanks!

PR #141, opened 2016-07-18
Don't rely on package version in README

asiniy:
As an elixir package maintainer I'm usually forgetting to update package version at README.md. So, please save tons of time to other developers by allow them to find their version by themselves ;)


alanpeabody:
Ha, I always forget to update this. Thanks!

Issue #143, opened 2016-07-23
Rendering bare resource objects


bcardarella:
I've been working on a json-api assertion library. I'd like to support ja_serializer but a few things are making it more difficult. Specifically how ja_serializer handles relationship data for resource objects.

Currently ja_serializer will raise an error if relationships aren't preloaded. I'd advocate for an option to skip this. Or perhaps not raise at all?

alanpeabody:
@bcardarella, just found that library the other day, excited to try it out and clean up a lot of test verbosity.

I am open to tweaking relationships, I have currently defaulted to behaving like ecto (eg no eager loading). The risk of not raising by default is people confused around missing relationships. I could see an option to skip if not loaded. What do you think a good API for that might look like?

 I have also been toying with a preload/2 hook that would get fired for either the collection or single struct being serialized. It would be a place to call Repo.preload, but only called on the top level collection or struct being serialized.

bcardarella:
Ideally all we'd need is a way to force the serializer to return a bare resource object. If you are open to it I can add support for a `bare: true` option on the context which will prevent relationships and includes from being rendered.

If this feels like a hack then I'm happy to do a hangout this week to discuss a better solution.

PR #144, opened 2016-07-24
Changes

bcardarella:
Please consider this a WIP branch. It does address #143 as well as some other blockers I've hit.

The 2nd change may appear to be subjective but I'd like to discuss the merits of it before any rejection is given.

1. Added `bare: true` to allow `JaSerializer.Builder.ResourceObject` to
build a bare object without requiring relationships. This can probably
be expected to `meta` and `links` as well. TBD.
2. Changed `JaSerializer.Formatter` to always output all keys as
strings. Currently the output format was a mix of atoms and strings.
Having a consistent key format will make payloads easier to test.
However, this may have an impact on the performance of the library and
this should be considered a breaking API change.

### DIFF
--- a/lib/ja_serializer/builder/relationship.ex
+++ b/lib/ja_serializer/builder/relationship.ex
@@ -6,8 +6,11 @@ defmodule JaSerializer.Builder.Relationship do
 
   defstruct [:name, :links, :data, :meta]
 
-  def build(%{serializer: serializer, data: data, conn: conn} = context) do
-    Enum.map serializer.relationships(data, conn), &(build(&1, context))
+  def build(%{serializer: serializer, data: data, conn: conn, opts: opts} = context) do
+    case opts[:relationships] do
+      false -> []
+      _ -> Enum.map serializer.relationships(data, conn), &(build(&1, context))
+    end
   end
 
   def build({name, definition}, context) do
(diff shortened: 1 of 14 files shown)


alanpeabody:
I don't have any objections to this.

RE: new option - is there a better name to use for this? Maybe `:attributes_only` or something similar if we also apply it to meta, links etc?

RE: binaries vs atoms & binaries - I think a consistent data structure makes a lot of sense. I would actually assume this would have a positive performance impact if anything, right?

While the second change is technically backwards incompatible I am not actually too worried about it because I assume 99% of the time results are passed immediately to a json encoder. I could be making a big assumption, but my guess is most people don't use or test the results directly.

bcardarella:
> is there a better name to use for this

definitely, `attributes_only` would imply only the attributes would come back. We also need `id` and `type`. Open to alternatives

> positive performance impact

I would think using more strings creates more objects subject to GC. It may be minimal though

bcardarella:
The alternative is to go the opposite and have booleans to shut off each piece:

`meta: false, links: false, relationships: false`

alanpeabody:
Ah, yeah, I wasn't thinking about id and type. The per key approach makes sense to me, even if we start with just relationships.

You will probably also want to skip sideloading the included relationships.

beerlington:
My unsolicited vote goes for `relationships: false`. It's explicit and uses the same language that should already be familiar to people.

bcardarella:
:+1: I'll update the PR

bcardarella:
Updated with `relationships: false`. I believe this will avoid sideloading.

alanpeabody:
Cool, thanks Brian. Merging now, but going to try to get out a release in the next few days with one other addition.

bcardarella:
Awesome!

PR #145, opened 2016-07-26
Spike out preload/3 hook.

alanpeabody:
@beerlington really this could be in JaResource instead, neither place is perfect but I thought having it close to the relationship definition made sense at some point. It might be helpful for people not using JaResource here as well.

If this idea is not totally crazy I will work on some basic tests.

### DIFF
--- a/lib/ja_serializer/serializer.ex
+++ b/lib/ja_serializer/serializer.ex
@@ -10,6 +10,7 @@ defmodule JaSerializer.Serializer do
     * `relationships/2`- A map of `HasMany` and `HasOne` data structures.
     * `links/2` - A keyword list of any links pertaining to this struct.
     * `meta/2` - A map of any additional meta information to serialized.
+    * `preload/3` - A special callback that can be used to preload related data.
 
   A Serializer (or view) is typically one of the few places in an API where
   content and context are both present. To accomodate this each callback gets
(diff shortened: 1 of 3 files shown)


beerlington:
This looks good. I'm thinking we could abstract it a little more in JaResource, but for this library I think it makes perfect sense. Couple typos in the docs that I'll comment on.

lib/ja_serializer/serializer.ex
```diff
@@ -145,6 +146,29 @@ defmodule JaSerializer.Serializer do
...
   defcallback links(map, Plug.Conn.t) :: map
 
+  @doc """
+  A special callback that can be used to preload related data.
+
+  Unlike the other callbacks, this callback is ONLY executed on the top level
+  data being serialized. Also unlike any other callback when serializing a list
+  of data (eg: from an index action) it recieves the entire list, not each
+  individual post. When serializing a single record (eg, show, create, update)
+  a single record is recieved.
```
hunterboerner: s/recieved/received

lib/ja_serializer/serializer.ex
```diff
@@ -145,6 +146,29 @@ defmodule JaSerializer.Serializer do
...
+
+  The primary usecase of the callback is to preload all the relationships you
```
hunterboerner: s/usecase/use case

alanpeabody:
Thanks @hunterboerner!

PR #147, opened 2016-08-04
Update MIME type configuration

coryodaniel:
Update MIME type configuration to reflect changes in Plug. Plug now uses Mime, and the config should be `:mime, :types`, instead of `:plug, :mimes`


alanpeabody:
Great, thanks!

PR #151, opened 2016-08-09
Deep merge relationship paths

### LINKED ISSUE #150: Parsed includes are not nested correctly, leading to missing resources in included
This could also be phrased as overlapping relationship paths not merged correctly.

The `opts[:include]` by the time it gets to `JaSerializer.Builder.Included.build/2` is `[rate_period_costs: [], job: [partner: [location: []]], job: [organization: []], job: [client: []]]`.  Notice how there are multiple keys entries for `job` because it's a `Keyword` list?  It's only picking up the first `:job` key, so it only sees `[:job][:partner][:location]` even though there is `:job :organization` and `:job :client`.  The original string includes are `"job-assignment.job.client,job-assignment.job.organization,job-assignment.job.partner.location,job-assignment.rate-period-costs"`.

I talked this over with @alanpeabody in Elixir Slack and the common keys (here `:job`) should be merged together (and potentially switching to a Map to prevent duplicate keys as happens with Keyword Lists).  I will work on a PR to fix this issue.  I'm opening it here, so other can find it.

KronicDeth:
Fixes #150

Keyword.merge was concatenating the nested lists instead of recursively merging the lists, so that relationships paths that shared relationships past the first level were never merged.  This then left the Keyword list with multiple copies of the same key and last code line Included.build only saw the first copy of the key (which was actually the last path in the string format because of the prepending to the normalized relationship path list).

### DIFF
--- a/lib/ja_serializer/builder/top_level.ex
+++ b/lib/ja_serializer/builder/top_level.ex
@@ -69,7 +69,7 @@ defmodule JaSerializer.Builder.TopLevel do
     normalized = path
     |> String.split(".")
     |> normalize_relationship_path
-    |> Keyword.merge(normalized, fn(_k, v1, v2) -> v1 ++ v2 end)
+    |> deep_merge_relationship_paths(normalized)
 
     normalize_relationship_path_list(paths, normalized)
   end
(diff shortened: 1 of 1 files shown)


alanpeabody:
Awesome thanks!

PR #159, opened 2016-08-17
Changed parse_realtionships function for current work with a nil realtionship

kaermorchen:
Hello.

When I change a realtionship id from null to 1 or from 1 to 2 it work as expected. But when I change a realtionship id from 1 to null, it work wrong - a nullable variable is excluded from a changes field in result changeset and misses in SQL query.


alanpeabody:
Thanks!

PR #161, opened 2016-08-19
Support additional error fields in EctoErrorSerializer

nurugger07:
Passing options for additional error fields was not actually including those fields
in the error object returned. This fixes that and now the error object can include
the additional errors.

I've also included some documentation for the module.

### DIFF
--- a/lib/ja_serializer/ecto_error_serializer.ex
+++ b/lib/ja_serializer/ecto_error_serializer.ex
@@ -1,35 +1,71 @@
 defmodule JaSerializer.EctoErrorSerializer do
   alias JaSerializer.Formatter.Utils
 
-  def format(errors), do: format(errors, %{})
-  def format(errors, conn), do: format(errors, conn, [])
-  def format(%Ecto.Changeset{} = cs, c, o), do: format(cs.errors, c, o)
-  def format(errors, _conn, _) do
+  @moduledoc """
+
+  The EctoErrorSerializer is used to transform Ecto changeset errors to JSON API standard
+  error objects.
+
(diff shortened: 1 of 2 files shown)


alanpeabody:
Great, thanks @nurugger07 !

wfgilman:
@nurugger07 There's a bit of an issue when using this with JaSerializer.PhoenixView. The options passed are accessed under the key `:opts` in `render_errors/1`. They are then put into `JaSerializer.EctoErrorSerializer.format/3` as the third argument, but not under the keyword `opts`. Therefore they are never found when `format/3` looks them up under the `opts` keyword and goes to merge them into the response. In order for it to work, the user needs to specify the `opts` in a nested fashion: `opts: [opts: [code: "DATABASE_ERROR]]`.

Here's an example using any old changeset.
```
JaSerializer.PhoenixView.render_errors(data: changeset, opts: [code: "DB_ERROR"])

%{"errors" => [%{detail: "Email This email has already been taken.",
     source: %{pointer: "/data/attributes/email"},
     title: "This email has already been taken."}],
  "jsonapi" => %{"version" => "1.0"}}
```
The additional field `code` never makes it in. The below works, however.
```
JaSerializer.PhoenixView.render_errors(data: changeset, opts: [opts: [code: "DB_ERROR"]])

%{"errors" => [%{code: "DB_ERROR",
     detail: "Email This email has already been taken.",
     source: %{pointer: "/data/attributes/email"},
     title: "This email has already been taken."}],
  "jsonapi" => %{"version" => "1.0"}}
```

PR #165, opened 2016-08-30
The example pagination links in the README (pasted below) fail due to…

MishaConway:
… the nil link and this is an attempt to fix that.

page = [
      first: "http://example.com/api/v1/posts?page[cursor]=1&page[per]=20",
      prev: nil,
      next: "http://example.com/api/v1/posts?page[cursor]=20&page[per]=20",
      last: "http://example.com/api/v1/posts?page[cursor]=60&page[per]=20"
    ]


alanpeabody:
Thanks!

PR #167, opened 2016-08-31
Fix crashes that occur when using custom pagination links

MishaConway:
containing a port number. 

The colon used for the port conflicts with param_fetcher_regex and the 'param' that is actually fetched in this case is the port. To get around this, I URI.parse the link. If it doesn't contain a port,
I let the existing implementation run. If it does, then I pass just the relative path portion of the link to the existing logic and with the result reassemble the fully qualified url.

### DIFF
--- a/lib/ja_serializer/builder/link.ex
+++ b/lib/ja_serializer/builder/link.ex
@@ -27,6 +27,25 @@ defmodule JaSerializer.Builder.Link do
   end
 
   defp path_for_context(context, path) do
+    uri = URI.parse(path)
+    path = uri
+    |> Map.put(:path, replaced_path_for_context(context, uri.path))
+    |> Map.put(:query, replaced_path_for_context(context, uri.query))
+    |> URI.to_string
+
+    # Remove trailing question mark if there is no query string.
+    # A query string itself can contain a trailing question mark so we only
+    # do this if URI did not parse out a query.
+    if nil == uri.query do
+      Regex.replace ~r/\?$/, path, ""
+    else
+      path
+    end
+  end
+
+  defp replaced_path_for_context(context, nil), do: ""
+
+  defp replaced_path_for_context(context, path) do
     @param_fetcher_regex
     |> Regex.replace(path, &frag_for_context(&1, context))
   end
--- a/test/ja_serializer/phoenix_view_test.exs
+++ b/test/ja_serializer/phoenix_view_test.exs
@@ -50,6 +50,15 @@ defmodule JaSerializer.PhoenixViewTest do
     assert Dict.has_key?(json, "links")
   end
 
+  test "render conn, index.json-api, model: model with custom pagination using urls with ports", c do
+    json = @view.render("index.json-api", conn: %{}, data: [c[:m1], c[:m2]],
+      opts: [page: [first: "http://localhost:4000/v1/posts/foo"]])
+    assert [a1, _a2] = json["data"]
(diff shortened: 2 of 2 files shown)


alanpeabody:
Great, thanks! Anyway you could add a test to make sure I don't accidentally break this in the future?

Good catch on the port issue btw.

MishaConway:
@alanpeabody Added a test and verified that before this change it catches the issue and that with this fix it passes.

lib/ja_serializer/builder/link.ex
```diff
@@ -27,6 +27,19 @@ defmodule JaSerializer.Builder.Link do
...
+    parsed_uri = URI.parse path
+    if parsed_uri.port do
+      replaced_path = replaced_path_for_context(context, parsed_uri.path)
+      absolute_replaced_path = "#{parsed_uri.scheme}://#{parsed_uri.host}:#{parsed_uri.port}#{replaced_path}"
+      [ absolute_replaced_path, parsed_uri.query ]
+      |> Enum.reject(&(is_nil(&1)))
+      |> Enum.join("?")
+    else
+      replaced_path_for_context(context, path)
+    end
```
alanpeabody: @MishaConway I am wondering if we should just always parse the ui and just run the replaced path on the path, something like this (untested):

```elixir
uri = URI.parse(path)
uri
|> Map.put(:path, replaced_path_for_context(context, uri.path))
|> Map.put(:query, replaced_path_for_context(context, uri.query))
|> URI.to_string
```

I think that may just cover all situations (except maybe using an atom in the url/domain/subdomain).

MishaConway:
@alanpeabody I think that will work, but will also have the side effect of relativizing any custom absolute urls a user explicitly provides. My assumption is that if absolute urls are provided then absolute urls should be rendered in the json, what is your opinion on that?

alanpeabody:
@MishaConway I would absolutely like to support both urls with hosts and without. My understanding was that `URI.parse`/`URI.to_string` would properly handle both cases properly, but I have only tried some basic experiments in iex:

```elixir
iex(1)> URI.parse("https://api.example.com/v1/foo/:id?bar_id=:id") |> IO.inspect |> URI.to_string
%URI{authority: "api.example.com", fragment: nil, host: "api.example.com",
 path: "/v1/foo/:id", port: 443, query: "bar_id=:id", scheme: "https",
 userinfo: nil}
"https://api.example.com/v1/foo/:id?bar_id=:id"
iex(2)> URI.parse("/v1/foo/:id?bar_id=:id") |> IO.inspect |> URI.to_string
%URI{authority: nil, fragment: nil, host: nil, path: "/v1/foo/:id", port: nil,
 query: "bar_id=:id", scheme: nil, userinfo: nil}
"/v1/foo/:id?bar_id=:id"
```

MishaConway:
@alanpeabody Sorry for the confusion, I relooked and think your suggestion will work and be a lot simpler. I started elixir last week and was also at work and briefly skimming, and I interpreted something different the first time I looked at your suggestion. 

I'll incorporate your suggestion and update this PR shortly.

MishaConway:
@alanpeabody I updated to use your suggestions with some slight modifications to handle nil values and urls with no query strings getting reassembled with trailing question marks (causing several tests to fail).

alanpeabody:
Awesome, thanks for finding and fixing the port issue @MishaConway! :100:

PR #181, opened 2016-10-01
Allow serialization of null structs

juanazam:
Hi,

First of all, thanks for this package, it's really great and has been very useful to me.

I have encountered a small problem while using it, let's say a have a resource (car for example) and another resource associated to it (insurance_data for example). The car may or may not have an insurance_data record associated to it and insurance_data records don't make sense if they are not associated to a car.

Given this reality I built a car serializer that has a belongs_to association to a insurance_data resource which (`link: [related: 'cars/:id/insurance_data']`).

I created and endpoint that given a car returns its insurance data:
`render(conn, "show.json-api", data: insurance_data)`.

When `insurance_data` is present, everything works as expected, the problem arises when the car doesn't have an insurance_data record associated to it.

`ja_serializer` is not allowing nil values to be serialized but the JSON API specification [allows it](http://jsonapi.org/format/#document-top-level).

```
Primary data MUST be either:

    a single resource object, a single resource identifier object, or null, for requests that target single resources
    an array of resource objects, an array of resource identifier objects, or an empty array ([]), for requests that target resource collections
```

This PR attempts to fix that behavior.

### DIFF
--- a/lib/ja_serializer.ex
+++ b/lib/ja_serializer.ex
@@ -55,8 +55,12 @@ defmodule JaSerializer do
   the conn and opts and returns a map ready for json encoding.
   """
   def format(serializer, data, conn \\ %{}, opts \\ []) do
-    %{data: data, conn: conn, serializer: serializer, opts: opts}
-    |> JaSerializer.Builder.build
-    |> JaSerializer.Formatter.format
+    cond do
+      data ->
+        %{data: data, conn: conn, serializer: serializer, opts: opts}
+        |> JaSerializer.Builder.build
+        |> JaSerializer.Formatter.format
+      is_nil(data) -> %{data: nil}
+    end
   end
 end
--- a/lib/ja_serializer/phoenix_view.ex
+++ b/lib/ja_serializer/phoenix_view.ex
@@ -110,21 +110,20 @@ defmodule JaSerializer.PhoenixView do
   end
 
   defp find_struct(serializer, data) do
-    case data[:data] do
-      nil ->
-        singular = singular_type(serializer.type)
-        plural = plural_type(serializer.type)
+    singular = singular_type(serializer.type)
+    plural = plural_type(serializer.type)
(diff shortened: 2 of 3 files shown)


alanpeabody:
TBH I don't really understand this use case. Would you not want to return a 404 if there is no model?

If not you can always opt out of using the ja_serializer for a particular request. I am wary of returning nil unexpectedly for people.

juanazam:
@alanpeabody looking at the JSON API spec (http://jsonapi.org/format/#fetching-resources-responses-200):

```
A server MUST respond to a successful request to fetch an individual resource with a resource object or null provided as the response document’s primary data.

null is only an appropriate response when the requested URL is one that might correspond to a single resource, but doesn’t currently.
```
The second sentence described my current situation. I'm using ember data and since returning 404 is not spec compliant, it is throwing an exception.

juanazam:
@alanpeabody 

> 
If not you can always opt out of using the ja_serializer for a particular request. I am wary of returning nil unexpectedly for people.

I forgot to mention I'm doing exactly that as a workaround but after reading the JSON API spec I thought it might be worth discussing it here.

yordis:
@alanpeabody yeah reading from the documentation 

> A server MUST respond to a successful request to fetch an individual resource with a resource object or null provided as the response document’s primary data.
> 
> null is only an appropriate response when the requested URL is one that might correspond to a single resource, but doesn’t currently.

but also

> A server MUST respond with 404 Not Found when processing a request to fetch a single resource that does not exist, except when the request warrants a 200 OK response with null as the primary data (as described above).

sulphur:
yes but having lots of 404 in js console for example is kind of ugly ;) especially when null struct is a normal respose like your example un empty cart or whatever relation that can be null in your database. so being able to render null is actually quite useful.

beerlington:
In reading the discussion and the spec, I agree that we should allow serializing nil values. The server can still choose to return a 404 if it wants, but this at least gives people the option to follow the spec where desired.

PR #185, opened 2016-10-11
[mix] Fixing generator task in umbrella apps;

gullitmiranda:
Small fix of the paths of the templates to be used by the generators


alanpeabody:
@gullitmiranda this still works on regular projects?

gullitmiranda:
yes. it works the same way as before

Issue #187, opened 2016-10-17
Raising error when trying to serialize an Ecto Error of {:array, :integer}


henriquecf:
Hello there, when I make a request with invalid data to a model that has an {:array, :integer} field, I get an error, that I suppose that happens when the error message is being serialized to be sent back to the client.

Here is the error:

` ** (UndefinedFunctionError) function String.Chars.Tuple.to_string/1 is undefined (module String.Chars.Tuple is not available)
        String.Chars.Tuple.to_string({:array, :integer})
        lib/ja_serializer/ecto_error_serializer.ex:45: anonymous fn/2 in JaSerializer.EctoErrorSerializer.format_each/2
        (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
        lib/ja_serializer/ecto_error_serializer.ex:44: JaSerializer.EctoErrorSerializer.format_each/2
        (elixir) lib/enum.ex:1184: Enum."-map/2-lists^map/1-0-"/2
        lib/ja_serializer/ecto_error_serializer.ex:37: JaSerializer.EctoErrorSerializer.format/2
        (phoenix) lib/phoenix/view.ex:335: Phoenix.View.render_to_iodata/3
        (phoenix) lib/phoenix/controller.ex:642: Phoenix.Controller.do_render/4`

Would you please help guide through solving this? I may send a pull request if I know where to correct this...

Thanks

alanpeabody:
Hi @henriquecf, it looks like we don't support whatever error message ecto is returning there.

It would just be adding a `format_each/2` definition that handled whatever structure your error is in. Code is at https://github.com/AgilionApps/ja_serializer/blob/master/lib/ja_serializer/ecto_error_serializer.ex#L42-L61 if you want to give it a shot.

`IO.inspect`'ing your invalid changeset.errors to figure out what the error looks like is the piece of data we will need to solve the issue.

henriquecf:
Inspecting the signature of format_each, I realized that its first received parameter is {:varying_amounts, {"is invalid", [type: {:array, :integer}]}}

The code that is generating the exception is the the reducer that generates the title on https://github.com/AgilionApps/ja_serializer/blob/master/lib/ja_serializer/ecto_error_serializer.ex#L44-L46
The error happens when the type {:array, :integer} is passed to the to_string method on line 45, but this field type is a valid changeset type on Ecto, so I am not sure what to change to make it work and not break other cases...

[...]

PR #190, opened 2016-10-21
Remove :type key when changing title on format_each

henriquecf:
This pull request was made based on discussion on #187


alanpeabody:
great, thanks!

PR #192, opened 2016-11-04
Fix #191, updated has_one to be :when_included by default

### LINKED ISSUE #191: Docs say has_one defaults to :when_included, but defaults to :always

zkwentz:


zkwentz:
Bumpy boo

alanpeabody:
Hey @zkwentz sorry, missed this in the flurry of notifications.

I kind of think we should update this in the docs instead of changing the current behavior, however I am open to feedback on why to go this direction.

beerlington:
It looks like we're defaulting to `:when_included` for a has_many relationship and to `:always` for a has_one. I'm assuming that's for performance reasons so we're not adding an extra querying for the related records by default. I'm going to go ahead an update the docs since that behavior makes the most sense to me. I'm happy to discuss further if anyone wants, but am going to close this ticket in favor of updating the docs.

PR #194, opened 2016-11-17
Allow pagination link keys to be configured

scrogson:
This PR allows the pagination links to be configured as desired, but maintains the original
as defaults.


alanpeabody:
Sweet, thanks!

PR #198, opened 2016-11-30
Fixes typos in EctoErrorSerializer

nmcalabroso:
Fixes all the typos where "then" was put instead of "than"


alanpeabody:
Thanks!

PR #200, opened 2016-12-13
Allow character encoding or other attributes in content-type header

sinakarimi:
I'm going to leave this pull request here so I can initiate a discussion about this. We came across an issue where an Android app (built with React Native) was trying to integrate with our API, however it seems that Android attaches character encoding properties to the content-type header e.g. `application/vnd.api+json; charset=utf-8`.

I see that in your tests you've defined `application/vnd.api+json; charset=utf-8` as an invalid content type, is there a reason behind this?

### DIFF
--- a/lib/ja_serializer/content_type_negotiation.ex
+++ b/lib/ja_serializer/content_type_negotiation.ex
@@ -32,7 +32,7 @@ defmodule JaSerializer.ContentTypeNegotiation do
   def verify_content_type(%Plug.Conn{method: "GET"} = conn, _o), do: conn
   def verify_content_type(%Plug.Conn{method: "DELETE"} = conn, _o), do: conn
   def verify_content_type(%Plug.Conn{} = conn, _o) do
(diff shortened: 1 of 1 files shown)


sinakarimi:
I can update the tests to make them pass if you're happy with this approach

alanpeabody:
@sinakarimi This is because per my understanding of JSON-API content negotiation that content-type header is invalid.

See http://jsonapi.org/format/#content-negotiation-servers and in particular:

> Servers MUST respond with a 415 Unsupported Media Type status code if a request specifies the  > header Content-Type: application/vnd.api+json with any media type parameters.

My understanding is basically that anything after a semi-colin is reserved for future use and should be rejected.

If this is insurmountable in your android app I suggest you don't use the provided content negotiation plug and write something custom that meets your needs.

sinakarimi:
Ok that's a valid point, I'll close this PR and create a custom content negotiation plug.

lleger:
Thanks @sinakarimi, I had the exact same issue and using your code in a custom Plug fixed the issue for me!

Issue #202, opened 2016-12-16
Meta on has many relationships.


zachdaniel:
Hey guys!  I really need to find a way to include metadata on the resource identifiers for many to many relationships.  It would be really great, if there was a way to return things in the form of `{record, meta}` from the `JaSerializer.DSL.has_many` function for this purpose.  Any thoughts?  I will need to hack together something soon if there isn't already a way to do it, so I may make a PR for this depending on your thoughts.

beerlington:
This no longer seems relevant because the associated PR was closed https://github.com/vt-elixir/ja_serializer/pull/203

PR #203, opened 2016-12-17
Allowing meta on relationships.  See issue #202

zachdaniel:
Haven't written any tests for this yet.  Didn't want to invest too much w/o feedback.

#202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## v0.11.2
+* Features
+  * Allow customization of pagination params in Scrivener intergration (@scrogson)
+* Bugfixes
+  * Add json api version to error responses. (@nmcalabroso)
+  * Support umbrella apps in generators. (@gullitmiranda)
+  * Support {:array, :integer} type ecto fields in error serializer	(@henriquecf)
+
+
 ## v0.11.1
 * Bugfixes
   * Handle port numbers in link urls. (@MishaConway)
--- a/README.md
+++ b/README.md
@@ -13,10 +13,6 @@ libraries such as Poison.
 
 Please open an issue or message/mention @alanpeabody in the [Elixir Slack](https://elixir-slackin.herokuapp.com/).
 
-## Shameless Plug
-
-JaSerializer development is sponsored by [Agilion](http://www.agilion.com).
-
 ## Usage
 
 See [documentation](http://hexdocs.pm/ja_serializer/) on hexdoc for full
--- a/mix.exs
+++ b/mix.exs
@@ -3,12 +3,12 @@ defmodule JaSerializer.Mixfile do
 
   def project do
     [app: :ja_serializer,
-     version: "0.11.1",
+     version: "0.11.2",
      elixir: "~> 1.1",
      build_embedded: Mix.env == :prod,
      start_permanent: Mix.env == :prod,
      consolidate_protocols: Mix.env != :test,
-     source_url: "https://github.com/AgilionApps/ja_serializer",
+     source_url: "https://github.com/vt-elixir/ja_serializer",
      package: package,
      description: description,
      deps: deps]
@@ -40,7 +40,7 @@ defmodule JaSerializer.Mixfile do
       licenses: ["Apache 2.0"],
       maintainers: ["Alan Peabody"],
       links: %{
-        "GitHub" => "https://github.com/AgilionApps/ja_serializer"
+        "GitHub" => "https://github.com/vt-elixir/ja_serializer"
       },
     ]
   end


alanpeabody:
Thanks @zachdaniel can you show some example usage? Per item meta on relationships is definitly a tough one.

zachdaniel:
On your `Person` serializer you might have something like this: 
```
  has_many :friends,
    serializer: PersonSerializer,
    identifiers: :when_included,
    meta_key: :friends,
    meta: [:inserted_at]
]
```
This represents the table `people` with a `friends` join table, that just has something like `person_1_id` and `person_2_id` and `inserted_at`.  I want to include in the meta for that resource identifier the time that they became friends.

What I'm doing currently (not all that great) is exposing the join table relationship on the other side, and selecting the one that corresponds to the individual record being serialized.  Kind of like this: 

```
def friends(person, conn) do
      friend_lookup =
      person.friends
      |> Enum.map(&({&1.person_2_id, &1}))
      |> Enum.into(%{})

    for friend <- person.friends do #friends was preloaded in the preload function
      join_table_row = friend_lookup[friend.id]
      %{friend| friends: [join_table_row]}
    end
end
```
So you're basically limiting their relationship to only reference the record that is currently being serialized.  Then, you put that on the relationship.  I tried to optimize this w/better preloads, but it doesn't work because you could serialize many Person's at one time, and so you can't isolate an individual one on the preload.

alanpeabody:
To be honest this doesn't really seem like what the meta property is for to me. This is describing real relationship data and probably belongs as an attribute or just fully representing both models in the UI. Typically this means it is information about the API (eg pagination, copyright, access information, etc) not the actual data model.

zachdaniel:
The problem is that this is data about a join table.  Where else could it go? Think of an example where you want many people, and their many friends.  How would you represent the time that all of those became friends?  I cant think of a way to do in a single request aside from metadata on the resource identifier.  It makes sense to do that as well because you would naturally get one relationship resource identifier for each join table row.

W/ JASON API lacking an extension system currently, metadata seems like it can and should be the release valve for this kind of pressure.

I can definitely see why you wouldn't want to condone something like this, but lacking some way to set the meta like that will just mean that we will have to maintain a fork.

alanpeabody:
@zachdaniel Honestly I would probably model the join table in the response as an include:

```elixir
# Person View

has_many :friendships,
  include: true
  serializer: FriendshipView

# Friendship View

attributes [:inserted_at]

belongs_to :friend,
  include: true,
  view: FriendView

# Friend View

attributes [:name, :etc]
```

Or if you really don't want that, make the join table the model you are actually rendering and just include the user attributes:

```elixir
# Person View

has_many :friends,
  include: true
  serializer: FriendView

def friends(person, _conn) do
  person.join_table_association_name
end

# Friend View

attributes [:inserted_at, :name, :user_inserted_at]

def name(friend, _), do: friend.user.name
def user_inserted_at(friend, _), do: friend.user.inserted_at
```

That said, I am very much for supporting meta data for relationships but it must be a flexible, generic API. In particular I would like to be able to support both relationship level meta (eg on the collection level) as well as the per resource identifier object case you are looking for.

zachdaniel:
For your second example though, I'm not looking for the inserted at of my friend, I'm looking for the inserted_at of the join table.  We have other fields on join tables that need to be rendered, and can't be represented as fields on one end or the other, because its a many to many relationships.

Modeling the join table as its own entity provides its own sets of challenges that reach out a lot further than the serialization layer.  We have a whole gamut of authorization/validation code that, so far, hasn't had to deal with join tables in terms of who can/can't see them and who can/can't alter them.  I think the *ideal* solution, after some perspective would be to just make a new callback, say in this case `friends_meta/3` which receives both ends of the relationship and the conn.  This way people can ignore those if they don't want them, but can use them if they do.  In my case, it would actually end up looking much better.

```
def friends_meta(person_1, person_2, _conn) do
  join_table_row =  Enum.find(person_1.friends, fn join_table -> join_table.person_2_id = person_2.id)
  %{
    inserted_at: join_table_row.inserted_at
  }
end
```

What do you think about something like that?

alanpeabody:
In the second example I was rendering the inserted at from both the join table and the person/user/whatever table as different names.

TBH, I think you might just be better off doing something custom as your requirements are very specific. I will check out any generic solutions you can come up with though.

Maybe something like:

```elixir
has_many :friends,
  include: true,
  serializer: FriendSerializer,
  meta: :friends_meta, # atom to function w/ arity 2
  resource_indentifier_meta: :join_table_meta # atom to function w/ arity 3

# Relationship level
def friends_meta(_person, _conn), do: %{description: "your friends!"}

# Per resource item
def join_table_meta(_this_person, _conn, _related_person), do: %{your_thing: "here"}

```

zachdaniel:
Oh, yeah I see now.  Thanks :)  I wasn't thinking of the separation between relationship level meta and identifier meta.  I'll implement what you mentioned most recently, and reopen that pr.

PR #205, opened 2016-12-19
Fix docs, improve readme.

pinx:
Added an example for PhoenixView, that shows how to handle relationships.


alanpeabody:
Awesome, thanks @pinx ! 🎆

PR #207, opened 2016-12-20
Allow configuration for global Scrivener base_url

geolessel:
I've found myself wishing I could set a `base_url` once and for all so that each Scrivener-generated pagination links section would use it without me having to pass in the option again and again. With that in mind, I make this proposal.

In order to set a global base_url for all links that don't have their own option passed in, configure the application:

```elixir
config :ja_serializer,
  scrivener_base_url: "http://example.com:4000/v1/"
```

Alternatively, I thought of implementing an option that would allow absolute URLs instead of relative. Something perhaps like this:

```elixir
config :ja_serializer, scrivener_base_url: :absolute # defaults to :relative
```

It would then construct the absolute host/port combo from the `conn` and append the relative path.

The benefit of the implemented method is the ability to specify exactly where the links should head. The benefit of the secondary method is most of the time the links should head to the current host/port combo and thus less `env`-specific configuration is necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,11 +345,19 @@ option.
 render conn, data: page, opts: [page: [base_url: "http://example.com/foos"]]
 ```
 
+You can also configure `ja_serializer` to use a global default URL
+base for all links.
+
+```elixir
+config :ja_serializer,
+  scrivener_base_url: "http://example.com:4000/v1/"
+```
+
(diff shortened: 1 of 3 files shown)


alanpeabody:
This makes sense and LGTM as is, thanks! 👍

PR #213, opened 2017-01-13
Fix Warnings For Elixir 1.4

asummers:
This fixes some of the compiler warnings present, in particular the ones that print every compilation from mix.exs. Strictly speaking, only the first commit is necessary, as that removes the most obnoxious warnings, but I figured I would just add parens everywhere. 

While going through and doing this, I noticed lots of Dict and HashSet deprecation warnings. Is this something that is done for compatibility purposes or has just gotten fixed yet?


alanpeabody:
Awesome, thanks @asummers!

I think at this point we could probably get rid of the Dict and HashSet usage. We dropped 1.1 from travis a while back and it might already be broken, a PR for that would be very welcome.

PR #216, opened 2017-01-20
Update poison to use >= 1.4.0, so that the latest poison can be used

### LINKED ISSUE #215: Poison dependency out of date
The current version of poison is 3.0.  Updating to allow for 3.0 will ease integration.

joshuataylor:
This also updates the other dependencies to their latest versions as
well.

Closes #215.

Let me know if `>=` is the correct syntax to use, as: `~> 1.4 or ~> 2.0 or ~> 3.0` doesn't seem to work.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -23,7 +23,7 @@ defmodule JaSerializer.Mixfile do
     [
       {:inflex, "~> 1.4"},
       {:plug, "> 1.0.0"},
-      {:poison, "~> 1.4 or ~> 2.0"},
+      {:poison, ">= 1.4.0"},
       {:ecto, "~> 1.1 or ~> 2.0", only: :test},
       {:earmark, "~> 0.1", only: :dev},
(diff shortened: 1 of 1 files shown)


alanpeabody:
Thanks @joshuataylor, TBH I am sure we could drop 1.x support for poison without problems, but this is great for now!

alanpeabody:
Actually, it looks like bumping ecto broke the elixir 1.2 test suite, anyway you could revert that update to the lock file (should still work for users)?

joshuataylor:
Okay, I removed the mix.lock changes, as this should be in a followup issue :)

I then ran `mix deps.update poison`, which didn't change the lock file:
```
Running dependency resolution...
Dependency resolution completed:
  poison 1.5.2
```

So I'm guessing that if you rely on poison 3.0 or above, it should work fine.

alanpeabody:
Yup, wonderful, thanks!

PR #226, opened 2017-02-21
Fix README example for pluralize_types config

petehamilton:
The example was incorrect compared to implementation


petehamilton:
@bcardarella thanks so much for adding this, was about to do so myself before noticing it'd been added in the last 24h! Had a confusing moment before checking the code and realising the example was wrong so just submitting a quick fix.

alanpeabody:
Thanks!

Issue #229, opened 2017-03-11
Multi-word relationships and inclusion behavior


lleger:
Let's say I have a resource with a relationship like so:

```elixir
has_many :author_comments
    serializer: MyApp.AuthorCommentView,
    include: false,
    identifiers: :when_included
```

And I make a request to include this: `?include=author-comments`

When I pass `params["include"]` to the render function, it's not rendering the relationship. There seems to be a disconnect between the underscored definition and the dasherized query param.

I was able to work around this by changing the definition and using `field`:

```elixir
has_many :"author-comments"
    serializer: MyApp.AuthorCommentView,
    include: false,
    identifiers: :when_included,
    field: :author_comments
```

But this feels wrong to me. Is there an error in my set up or is this a bug in JaSerializer?

alanpeabody:
@yogipatel pushed a fix for this, thanks for the report!

PR #230, opened 2017-03-18
Parse/format included relationship key paths

yogipatel:
Addresses #229 

Previously, if a relationship contained multiple words (`has_many :author_comments`) the client would be forced to send `include=author_comments` instead of the expected `author-comments`.

This PR addresses this by utilizing `JaSerializer.ParamParser.Utils.format_key/1`'s format conversion.


alanpeabody:
Thanks @yogipatel ! ❤️

PR #233, opened 2017-03-20
[Proposal] Add JaSerializer.Builder.PaginationLinks

### LINKED ISSUE #231: Pagination offset and limit values harcoded
The pagination rules for Scrivener make an assumption of `number` and `size` for the pagination names. However, the json-api spec notes that other pagination strategies can be used like `offset` and `limit` which would be a zero-based index value. It seems the current implementation wouldn't support this value despite the key names being configurable.

bcardarella:
**This introduces some breaking changes**

Closes #231 

Introduces breaking changes in the form of the page keys.

I opted to normalize on the suggested `page[number]` and `page[size]`
nomenclature that is used as an example in the json-api `page` spec:

http://jsonapi.org/format/#fetching-pagination

Keep in mind that this nomenclature is not the "official" one, as there
isn't one but in my experience this is a very common implementation over
the original `page[page]` and `page[page-size]`

This PR also has the `ScrivenerLinks` builder consume `PaginationLinks`
builder.

For the time being I suggest building links and passing to the `page`
option for the view which is why this builder's API does not conform to
the same API as the other builders.

This PR also give configuration control over the page number origin
value. Previously this was hardcoded to `1` but now you can override if
you prefer the `offset/limit` nomenclature.

Breaking changes:

* `page_number_key` defaults to `number` instead of `page`
* `page_size_key` defaults to `size` instead of `page-size`
* `scrivener_base_url` renamed to `page_base_url`


lib/ja_serializer/builder/pagination_links.ex
```diff
@@ -0,0 +1,94 @@
+defmodule JaSerializer.Builder.PaginationLinks do
+  import JaSerializer.Formatter.Utils, only: [format_key: 1]
+
+  @page_number_origin Application.get_env(:ja_serializer, :page_number_origin, 1)
```
bcardarella: I'll update the README. If this value changes a recompile of ja_serializer will have to be forced

bcardarella:
@alanpeabody any thoughts on this?

alanpeabody:
@bcardarella Sorry, haven't had an opportunity to review yet, but thanks for the thought and effort you have put into it. I will make sure to take a look tonight.

bcardarella:
Thanks! 🎉

alanpeabody:
@bcardarella Finally getting a chance to sit down and look at this.

I think the only thing I am concerned with is breaking backwards compatibility. I guess It would also be nice to support cursor based pagination, but that might just need to be custom anyway.

I do think the pagination names you suggest are much better (and less lazy then the scrivener defaults), and the more I think about it I think the pagination changes would be super simple for people to make when updating.

Talked myself into it.

alanpeabody:
Thanks @bcardarella !

bcardarella:
thanks!

PR #250, opened 2017-08-01
add render to overridable

atomkirk:
looking for feedback


alanpeabody:
@atomkirk, TBH, if you don't want render you can probably just not use the JaSerializer.PhoenixView.

In your normal view all you need to do is define your own render that calls JaSerializer:

```elixir
defmodule MyApp.BlogView do
  use MyApp.Web, :view
  use JaSerializer #(use the base serializer)

  def render("index.json-api", %{blogs: blogs, conn: conn}) do
    # whatever custom logic you wanted to do.
    JaSerializer.format(__MODULE__, blogs, conn, [])
  end

  # dsl/or callbacks ...
end

PR #252, opened 2017-08-18
Fix an edge case with pagination links

marpo60:
* When page_key is nil and the conn query params has page information
  page_url  was not overriding the page params

  For example:

  original_params contains `%{"number" => 3}`

  `%{page_key() => %{page_number_key() => number, page_size_key() =>
  %size}}` evalutes to `%{nil => %{"number" => 1, "size" => 20}`

  So when merging original_params with the above the map ends up with

`%{"number" => 3, nil => %{"number" => 1, "size" => 20}}`

  And the Plug.Conn.Query.encode of that map evalutes to
  "number=1&size=20&number=3"

### DIFF
--- a/lib/ja_serializer/builder/pagination_links.ex
+++ b/lib/ja_serializer/builder/pagination_links.ex
@@ -74,10 +74,17 @@ defmodule JaSerializer.Builder.PaginationLinks do
   defp page_url(number, base, size, orginal_params) do
     params =
       orginal_params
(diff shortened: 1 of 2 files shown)


alanpeabody:
I kind of feel like this is outside the specification:

http://jsonapi.org/format/#fetching-pagination

Although it is a "SHOULD", not a "MUST".

Is there a particular reason you are not using the page param?

PR #254, opened 2017-09-13
Params: Preserve ID in to_attributes/1

xtian:
This PR changes Params.to_attributes/1 to keep the id of the resource in the same way as the type is preserved. Is there a reason it was left out?


alanpeabody:
Honestly I don't remember why it was left out, but I don't see an issue with this.

xtian:
Thanks!

PR #255, opened 2017-10-11
Fix relationship key formatting for multiword relationship names

jerel:
Currently any multi-word relationship names will end up formatted as `user-account_id` so this PR builds the id and then formats it.

### DIFF
--- a/lib/ja_serializer/params.ex
+++ b/lib/ja_serializer/params.ex
@@ -1,4 +1,5 @@
 defmodule JaSerializer.Params do
+  import JaSerializer.ParamParser.Utils, only: [format_key: 1]
   @moduledoc """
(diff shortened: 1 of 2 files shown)


jerel:
You're right. I was using `JaSerializer.Params.to_attributes/1` in a channel and the unexpected behavior was because the `JaSerializer.ParamParser.parse/1` plug was not in play. I combined `parse/1` and `to_attributes/1` into a custom `to_attributes/1` and it works without my "fix".

By the way, is another Hex release likely to happen soon? We've been running off of a git hash because the last release was in February

beerlington:
@jerel I'm hoping to get a new release out in the next couple weeks. Stay tuned!

PR #256, opened 2017-10-12
Add identifiers option

raelity:
Specifying an `identifiers` option overrides the identifiers field, including identifiers for the relationships specified, allowing the inclusion of specific relationship identifiers independent of what's included (with `:when_included`) and without need of always including identifiers (with `:always`).


raelity:
Typo'd on origin — should have been against my own fork.

PR #258, opened 2017-11-07
Discard non-map items when building includes WIP

cjfreeze:
If we want to have the builder create a resource identifier for non-existent includes, we need to provide the ID of the non-existent item when we receive nil in our serialize. But `included.ex` relies on the relationships being maps. To fix this we filter any non-maps out which prevents them from appearing in the included json object, making the behavior the same as if it were not included at all.


beerlington:
@cjfreeze thanks for the contribution. Any chance you'd be able to add a test to demonstrate the issue?

cjfreeze:
@beerlington Yes sorry for not labeling this a Work In Progress.

PR #264, opened 2017-12-13
Sparse fields should consider relationships as field

### LINKED ISSUE #240: Sparse fields should consider relationships as field
Currently with the following code:
```
has_many :comments, serializer: MyApp.CommentView, include: false, identifiers: :when_included
```

And requesting with sparse fields: `http://localhost:4000/posts?fields[posts]=name` the resulting response has `attributes` correctly filtered but has the following additional fields:
```
# ...
"relationships": {
   "comments": {}
}
# ...
```

And when using `identifiers: :always` the response also have the following additional fields:
```
# ...
"relationships": {
  "comments": {
    "data": [
      {
        "type": "S3FileSet",
        "id": "33768c8a-a7e7-448e-ad2c-4279228b5bf4"
      }
  }
}
# ...
```

I think for both the above cases the response should respect the sparse fields, and NOT include the `comments` or even better to just omit the entire `relationships` object, since the client is only requesting the `name` fields which is an attribute and in the JSON API spec it states:

> If a client requests a restricted set of __fields__ for a given resource type, an endpoint MUST NOT include additional fields in resource objects of that type in its response.

The definition for **fields** in the JSON API spec include both attributes and relationships:
> A resource object’s **attributes** and its **relationships** are collectively called its “fields”.

but right now only attributes are being filtered, relationships are not

marpo60:
Closes #240

### DIFF
--- a/lib/ja_serializer/builder/relationship.ex
+++ b/lib/ja_serializer/builder/relationship.ex
@@ -9,7 +9,10 @@ defmodule JaSerializer.Builder.Relationship do
   def build(%{serializer: serializer, data: data, conn: conn, opts: opts} = context) do
     case opts[:relationships] do
       false -> []
-      _ -> Enum.map serializer.relationships(data, conn), &(build(&1, context))
+      _ ->
+        serializer.relationships(data, conn)
+        |> filter_fields(context)
+        |> Enum.map(&(build(&1, context)))
     end
   end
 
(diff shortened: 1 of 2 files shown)


beerlington:
Closing in favor of https://github.com/vt-elixir/ja_serializer/pull/324

Issue #261, opened 2017-11-24
Different key_format in dev and prod environment


AxelTheGerman:
Hi,
I noticed that in my production environment the key_format is the default `:dasherized` instead of my configured `:underscored`.

here are some code snippets:
`config/config.exs`
```
config :mime, :types, %{
  "application/vnd.api+json" => ["json-api"]
}

config :ja_serializer,
  key_format: :underscored

# Import environment specific config. This must remain at the bottom
# of this file so it overrides the configuration defined above.
import_config "#{Mix.env}.exs"
```
`config/dev.exs`
```
config :sauron, MyApp.Endpoint,
  http: [port: 4000],

  # debug_errors must be false in order for sauron_client gem to parse error cases.
  # set `debug_errors: true` on your local machine if you need the interactive debugger.
  debug_errors: false,
  code_reloader: true,
  cache_static_lookup: false,
  check_origin: false,
  watchers: []

# Watch static and templates for browser reloading.
config :sauron, MyApp.Endpoint,
  live_reload: [
    patterns: [
      ~r{priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$},
      ~r{web/views/.*(ex)$},
      ~r{web/templates/.*(eex)$}
    ]
  ]
```

`config/prod.exs`
```
import_config "airbrake.exs"

config :sauron, MyApp.Endpoint,
  http: [port: 4000],
  url: [host: System.get_env("APP_HOST"), port: 80],
  cache_static_manifest: "priv/static/cache_manifest.json",
  secret_key_base: System.get_env("SECRET_KEY_BASE")
```

`web/router.exs`
```
defmodule MyApp.Router do
  use Airbrakex.Plug
  use MyApp.Web, :router

  pipeline :api do
    plug :put_format, "json-api"
  end
```

`web/views/api/v1/activity_view.ex`
```
defmodule Sauron.Api.V1.ActivityView do
  use JaSerializer.PhoenixView

  attributes ~w(
    source
    event_type
    action
    event_group
    request
    user
    organization
    primary_container
    secondary_container
    target
    record
    context_data
    logged_at
    sanitized_at
  )a
```

`mix.exs`
```
  def project do
    [app: :my_app,
     version: "0.0.1",
     elixir: "~> 1.5",
     elixirc_paths: elixirc_paths(Mix.env),
     compilers: [:phoenix] ++ Mix.compilers,
     build_embedded: Mix.env == :prod,
[...]

Issue #170, opened 2016-09-06
Custom key formatting when using scrivener broken.


edmaarcosta:
**Elixir 1.3.2**

I have a project using ja_serializer with scrivener_ecto for pagination, also a config for custom _dasherize_ in config.exs. But this message appears in `mix do deps.get, deps.compile`:

`== Compilation error on file lib/ja_serializer/builder/scrivener_links.ex ==
** (UndefinedFunctionError) function MyApp.Dasherize.dasherize/1 is undefined (module MyApp.Dasherize is not available)
    MyApp.Dasherize.dasherize("page")
    lib/ja_serializer/builder/scrivener_links.ex:8: (module)
`

If I comment the custom config, the compiling happens fine.

This is my config.exs:
```
config :ja_serializer, key_format: {:custom, MyApp.Dasherize, :dasherize}
```

This is my module for custom dasherize:
```
defmodule MyApp.Dasherize do
  def dasherize(key) do
    "__id__" ->
      key
    _ ->
      underscore_to_dash(key)
  end

  defp underscore_to_dash(key), do: String.replace(key, "_", "-")
end
```

This is my mix.exs
```
  defp deps do
    [
     {...},
     {:ja_serializer, "~> 0.10.1"},
     {:scrivener_ecto, "~> 1.0"}
    ]
  end
```

ja_serializer search for dasherize before the module being compiled.

The repo is [here](https://github.com/edmaarcosta/my_app) case required.

alanpeabody:
Thanks for the bug report w/ repo! I will take a look this evening.

alanpeabody:
Okay @edmaarcosta I understand the issue. It was introduced when working on improving performance. I will push out a new version with a fix in the next couple of days.

PR #265, opened 2018-01-11
Read key format config value at runtime

Frost:
Previously, the `:key_format` config value was read into a module attribute
`@key_formatter` in `JaSerializer.Formatter.Utils`.

It was therefore read from the config at compile-time and stored in the compiled
module.

This leads to two things:

1. Tests using `Application.put_env/3` don't work as expected
   The tests that use `Application.put_env/3` don't work as one might think they
   do, since while the `put_env` call does change the config value, it does not
   change the already compiled `@key_formatter` module attribute, which means
   that all the tests will keep using the default key formatting, regardless of
   the config value.

2. The library has to be recompiled after changing `key_format` config
   The library has to be recompiled at any time the `key_format` config key is
   changed, which has led to issues #63 and #261. #170, which was followed up by
   #172, fixes a similar issue.

I found [this article on erlang-solutions.com](https://www.erlang-solutions.com/blog/elixir-module-attributes-alchemy-101-part-1.html)
 that talks a bit more about module attributes and contains this nice piece of advice:

> The takeaway from this is: when using module attributes, always be aware of
> what they evaluate to, making a wrong assumption can result in having to
> recompile your project.

One other thing I fix here is that the custom key formatter set in one test in
included_tests.exs, `{:custom, Macro, nil, :underscore}` does not work, as it
ends up trying to call `Macro.nil`, which does not exist. Instead, I replaced
this with `{:custom, String, :capitalize, :downcase}` and changed any failing
tests accordingly.

### DIFF
--- a/test/ja_serializer/builder/included_test.exs
+++ b/test/ja_serializer/builder/included_test.exs
@@ -415,24 +415,25 @@ defmodule JaSerializer.Builder.IncludedTest do
 
     json =
       JaSerializer.format(ArticleSerializer, a1, %{},
-        include: "author.publishing-agent"
+        include: "author.publishing_agent"
       )
 
     ids = Enum.map(json["included"], &Map.get(&1, "id"))
-    refute "p1" in ids
+
+    assert "p1" in ids
 
     Application.put_env(
       :ja_serializer,
       :key_format,
-      {:custom, Macro, nil, :underscore}
+      {:custom, String, :capitalize, :downcase}
     )
 
     json =
       JaSerializer.format(ArticleSerializer, a1, %{},
-        include: "author.publishing-agent"
+        include: "Author.Publishing_agent"
       )
 
     ids = Enum.map(json["included"], &Map.get(&1, "id"))
-    refute "p1" in ids
+    assert "p1" in ids
   end
 end
--- a/README.md
+++ b/README.md
@@ -415,8 +415,6 @@ config :ja_serializer,
   key_format: {:custom, MyStringModule, :camelize, :underscore}
 ```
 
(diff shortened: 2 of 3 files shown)


Frost:
TL;DR on this: These changes remove the need to run `mix deps.clean ja_serializer && mix deps.get` after changing the `key_format` config.

test/ja_serializer/builder/included_test.exs
```diff
@@ -306,13 +306,13 @@ defmodule JaSerializer.Builder.IncludedTest do
     assert "p1" in ids
 
     Application.put_env(:ja_serializer, :key_format, :underscored)
-    json = JaSerializer.format(ArticleSerializer, a1, %{}, include: "author.publishing-agent")
+    json = JaSerializer.format(ArticleSerializer, a1, %{}, include: "author.publishing_agent")
     ids = Enum.map(json["included"], &(Map.get(&1, "id")))
-    assert not "p1" in ids
```
beerlington: Why is this assertion changing?

Frost: That is because I also changed the string `publishing-agent` to `publishing_agent` two lines above, in order to make a positive assertion to prove that the `:underscored` key format also works.

The same goes for the assertion a few lines down as well.

It might make sense to split this test into multiple different test cases in a `describe "multi-word relationship path keys"` block, and then also have both positive and negative assertions for both `:dasherized` and `:underscored` key formats.

Frost:
@beerlington This one is also rebased now, and I made sure to run `mix format` on the changed files again.

PR #276, opened 2018-05-04
[fix] ja_serializer#275 add support for Changeset.unsafe_validate_unique validation

sescobb27:
fixes https://github.com/vt-elixir/ja_serializer/issues/275


rtablada:
@beerlington can we get a patch release for this fix?

PR #281, opened 2018-06-22
fixes #279 build pagination URLs using base url and request path

### LINKED ISSUE #279: base url in pagination not using request path
Hi, i'm trying to paginate a bunch of resources using scrivener, but i'm seeing issues when using 
```ex
config :ja_serializer, page_base_url: "http://localhost:4200"
```
the problem is that i'm requesting `http://localhost:4200/api/places` but the links that `ja_serializer` with `scrivener` are generating didn't include the `/api/places` path, but instead is just returning `http://localhost:4200?page[number]=12&page[size]=10` i think the problem is here https://github.com/vt-elixir/ja_serializer/blob/master/lib/ja_serializer/builder/pagination_links.ex#L103 and also i think we should do this, instead:

```ex
defp base_url(conn, nil) do
  Application.get_env(:ja_serializer, :page_base_url, "") <> conn.request_path
end
defp base_url(_conn, url), do: url
```

with that, we can prepend the base url to the `conn.request_path` generating valid responses. `http://localhost:4200/api/places?page[number]=12&page[size]=10`

any thought?
thanks.

sescobb27:


beerlington:
Looks good, thanks!

PR #293, opened 2018-10-01
Handle the case where the end of a link is a URI fragment e.g. /foo#path

AlexRiedler:
The existing code would add a random question mark '?' inbetween
producing "/foo?#path" this is unnecessary and should become
"/foo#path". Instead of using a regex to remove, lets just not set it on
the URI if the query parameter is nil.

### DIFF
--- a/lib/ja_serializer/builder/link.ex
+++ b/lib/ja_serializer/builder/link.ex
@@ -29,21 +29,14 @@ defmodule JaSerializer.Builder.Link do
 
   defp path_for_context(context, path) do
     uri = URI.parse(path)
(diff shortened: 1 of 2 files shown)


beerlington:
@AlexRiedler I'm finally getting around to reviewing all the existing PRs and this looks good. I pushed an update to master last night that formats the code using `mix format` so this will need to be rebased. If you're able to resolve the merge conflicts, I can get this merged in.

AlexRiedler:
@beerlington thanks! I have rebased the changes :)

beerlington:
Thanks!

PR #294, opened 2018-10-12
Added missing @callback in Serializer

lasseebert:
Added missing @callback for `type/0` in Serializer

When this callback is missing, the calling code can't mark the type/0 function as a behaviour implementation like so:

```elixir
@impl JaSerializer.Serializer
def type, do: "category"
```


beerlington:
Thanks!

PR #297, opened 2018-11-21
Only include attributes key when there are attributes present

moxley:
We had a schema validation error in our endpoint responses, because the Formatter added an `attributes` key to the response, but our resource doesn't have attributes (it only has relationships).

### DIFF
--- a/lib/ja_serializer/formatter/resource.ex
+++ b/lib/ja_serializer/formatter/resource.ex
@@ -6,12 +6,12 @@ defimpl JaSerializer.Formatter, for: JaSerializer.Builder.ResourceObject do
     links = Utils.array_to_hash(resource.links)
 
     json = %{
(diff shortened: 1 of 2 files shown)


beerlington:
Thanks! can you please add a test to cover this case?

moxley:
@beerlington: Tests added. Thanks!

beerlington:
Thanks! hoping to get a release out soon

jeregrine:
This is super old but https://jsonapi.org/format/#document-resource-object-attributes states that attributes MUST be present

beerlington:
@jeregrine I'm not sure about that. I see `The value of the attributes key MUST be an object` and [here](https://jsonapi.org/format/#document-resource-objects) I see:

Screen Shot 2020-10-19 at 8 12 30 AM

PR #299, opened 2018-11-29
Consistent relation override

### LINKED ISSUE #298: Inconsistency in attribute and relationship functions
In a view, one can specify a custom attribute in two ways:

```elixir
attributes [:foo]

def foo(model, _conn), do: model.bar
# OR:
def foo(model), do: model.bar
```

For relationships only the first way works:

```elixir
has_one(:foo, include: true, serializer: SomeView)

def foo(model, _conn), do: model.bar
```

This inconsistency caught me and I spend some time figuring out what was wrong. (Especially because the default behaviour just defaults to `nil` if a function or attribute is not found).

Is this intended? Otherwise I will happily make a PR making it consistent.

lasseebert:
Closes #298

For attributes it is possible to override function with either one or two arguments: `(model)` or `(model, conn)`.
For relations, only the `(model, conn)` override is possible. This is inconsistent and might confuse.

This PR adds the `(model)` override to relations too.

### DIFF
--- a/lib/ja_serializer/relationship.ex
+++ b/lib/ja_serializer/relationship.ex
@@ -137,10 +137,13 @@ defmodule JaSerializer.Relationship do
   @doc false
   def default_function(name, opts) do
     quote bind_quoted: [name: name, opts: opts] do
-      def unquote(name)(struct, _conn) do
+      def unquote(name)(struct) do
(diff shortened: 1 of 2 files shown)


beerlington:
Thanks!

lasseebert:
Thanks for the quick review and merge :)

PR #304, opened 2019-01-04
Fix compilation error when defining already inlined by compiler methods

marpo60:
Fixes https://github.com/vt-elixir/ja_serializer/issues/283
Related https://github.com/vt-elixir/ja_serializer/pull/245

Benchmark before the change:
```
Settings:
  duration:      1.0 s

## JaSerializer.PhoenixViewBench
[17:16:16] 1/6: attributes map big data
[17:16:21] 2/6: attributes map small data
[17:16:23] 3/6: render index 25 items map big data
[17:16:25] 4/6: render index 25 items small data
[17:16:29] 5/6: render index one item map big data
[17:16:32] 6/6: render index one item small data

Finished in 17.87 seconds

## JaSerializer.PhoenixViewBench
attributes map small data             10000000   0.12 µs/op
attributes map big data               10000000   0.46 µs/op
render index one item small data         20000   84.84 µs/op
render index one item map big data       20000   93.65 µs/op
render index 25 items small data         10000   292.75 µs/op
render index 25 items map big data        2000   820.56 µs/op
```

Benchmark after the change

```
Settings:
  duration:      1.0 s

## JaSerializer.PhoenixViewBench
[17:15:36] 1/6: attributes map big data
[17:15:42] 2/6: attributes map small data
[17:15:43] 3/6: render index 25 items map big data
[17:15:46] 4/6: render index 25 items small data
[17:15:48] 5/6: render index one item map big data
[17:15:50] 6/6: render index one item small data

Finished in 17.07 seconds

## JaSerializer.PhoenixViewBench
attributes map small data             10000000   0.13 µs/op
attributes map big data               10000000   0.53 µs/op
render index one item small data         20000   83.12 µs/op
render index one item map big data       20000   92.74 µs/op
render index 25 items small data          5000   288.62 µs/op
render index 25 items map big data        2000   785.51 µs/op
```

Best!

### DIFF
--- a/lib/ja_serializer/dsl.ex
+++ b/lib/ja_serializer/dsl.ex
@@ -88,18 +88,19 @@ defmodule JaSerializer.DSL do
   end
 
   defp define_inlined_attributes_map(env) do
-    attributes = Module.get_attribute(env.module, :attributes)
+    view = env.module
+    attributes = Module.get_attribute(view, :attributes)
     conn = quote do: conn
     struct = quote do: struct
 
-    # Construct ASL for map with keys from attributes calling the attribute fn
-    body =
-      {:%{}, [], Enum.map(attributes, fn k -> {k, {k, [], [struct, conn]}} end)}
-
     quote do
       @compile {:inline, inlined_attributes_map: 2}
-      def inlined_attributes_map(unquote(struct), unquote(conn)),
-        do: unquote(body)
+      def inlined_attributes_map(unquote(struct), unquote(conn)) do
+        # Construct ASL for map with keys from attributes calling the attribute fn
+        unquote({:%{}, [], Enum.map(attributes, fn k ->
+          {k, {{:., [], [{:__aliases__, [], [view]}, k]}, [], [struct, conn]}}
+        end)})
+      end
 
       defoverridable attributes: 2
     end
@@ -250,7 +251,7 @@ defmodule JaSerializer.DSL do
         @compile {:inline, [{att, 1}, {att, 2}]}
 
         def unquote(att)(m), do: Map.get(m, unquote(att))
-        def unquote(att)(m, c), do: unquote(att)(m)
+        def unquote(att)(m, c), do: __MODULE__.unquote(att)(m)
         defoverridable [{att, 2}, {att, 1}]
       end
     end
--- a/lib/ja_serializer/dsl.ex
+++ b/lib/ja_serializer/dsl.ex
@@ -97,9 +97,13 @@ defmodule JaSerializer.DSL do
       @compile {:inline, inlined_attributes_map: 2}
       def inlined_attributes_map(unquote(struct), unquote(conn)) do
         # Construct ASL for map with keys from attributes calling the attribute fn
-        unquote({:%{}, [], Enum.map(attributes, fn k ->
-          {k, {{:., [], [{:__aliases__, [], [view]}, k]}, [], [struct, conn]}}
-        end)})
+        unquote(
+          {:%{}, [],
+           Enum.map(attributes, fn k ->
+             {k,
+              {{:., [], [{:__aliases__, [], [view]}, k]}, [], [struct, conn]}}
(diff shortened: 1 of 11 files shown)


beerlington:
Awesome, thanks!

beerlington:
@marpo60 I just tried running this on my production app and am getting errors related to overriding attributes in the view like this:

```
defmodule MyView do
  use JaSerializer.PhoenixView

  attributes([:my_attr])

  def my_attr(%{my_attr: my_attr}), do: my_attr
end
```

produces...
```
function MyView.my_attr/1 is undefined or private
```

I haven't had a chance to dig in too much, but maybe there isn't test coverage around this or my app is doing something we're not testing. I might need to revert this if I can't figure out what's going on.

marpo60:
@beerlington I just try both cases on a local branch and they are working for me.
Im not able to reproduce the bug 😕

https://github.com/marpo60/ja_serializer/commit/2af4ee9a968a1e48ce4ba629d71e31ea1bb5318a

beerlington:
@marpo60 I dug in a little more and think I created a test that reproduces it. I want to spend a little more time with it this evening and see if it's something we're doing in our production app or not.

lib/ja_serializer/dsl.ex
```diff
@@ -250,7 +255,7 @@ defmodule JaSerializer.DSL do
         @compile {:inline, [{att, 1}, {att, 2}]}
 
         def unquote(att)(m), do: Map.get(m, unquote(att))
-        def unquote(att)(m, c), do: unquote(att)(m)
+        def unquote(att)(m, c), do: __MODULE__.unquote(att)(m)
```
beerlington: @marpo60 I figured out what the issue is. In my app, we have the view functions as private functions. So by changing this to `__MODULE__.unquote`, the private functions can no longer be called. I'm not sure if it's possible, but I would love to retain support for defining the attribute functions as private. 

I pushed up a test that reproduces the issue: https://github.com/vt-elixir/ja_serializer/compare/private-attribute-fn

Let me know if you're interested in fixing, otherwise I'll take a stab at it later this week.

beerlington: I dug in a little more and it looks like we only supported the single arity version as private, but the two arity functions had to be public. This was not documented and not officially supported, so I think we're good with your change. The single version just happened to support being private because of how it was being called, but this was inconsistent.

I'll add a note in the changelog about this being a potentially breaking change and that users just need to make their private functions public.

Thanks again!

marpo60: @beerlington I investigate a bit yesterday night and here are my findings.

I added this test:

```
  test "x" do
    defmodule PrivateArticle do
      use JaSerializer

      attributes([:private_arity_one, :private_with_two])

      defp private_arity_one(_a), do: "private"
      defp private_arity_two(_a, _b), do: "private"
    end

    article = %TestModel.Article{}
    assert PrivateArticle.attributes(article, %{}) == %{
      private_arity_one: "private",
      private_arity_two: "private"
    }
  end
```

I tested this test in this commit https://github.com/vt-elixir/ja_serializer/commit/b5d9f1d right before https://github.com/vt-elixir/ja_serializer/pull/245 got merged.

The test is not working.

```
  1) test x (JaSerializer.SerializerTest)
     test/ja_serializer/serializer_test.exs:41
     ** (UndefinedFunctionError) function JaSerializer.SerializerTest.PrivateArticle.private_arity_one/1 is undefined or private
```

I then tested it after that PR got merged https://github.com/vt-elixir/ja_serializer/commit/11e79da

The test is green.

IMHO this was just a side effect, and a not intended one caused by changing this:
https://github.com/vt-elixir/ja_serializer/pull/245/files#diff-bbdfd0bc4622bc60b9c795038912ad87L228

Having the definition of attributes as private function will conflict with inline by compiler methods, so I think it should be a must that if you overwrite an attribute it should with a public function.

Cheers.

PR #310, opened 2019-02-12
Use extra_applications over applications

### LINKED ISSUE #309: Can not start app without poison
After https://github.com/vt-elixir/ja_serializer/pull/307, it is no longer possible to use ja_serializer without poison, since it is still part of the `applications` (in mix.exs)[https://github.com/vt-elixir/ja_serializer/blob/084125c5733733871a51cc3a0d4a5d3957592aa3/mix.exs#L21]

I guess this could just be replaced with `extra_applications: [:logger]`.

Will make a PR ;)

lasseebert:
Closes #309 

This will not attempt to start poison app when in a non-test environment.
Otherwise, poison is needed to use ja_serializer as a dependency.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -18,7 +18,7 @@ defmodule JaSerializer.Mixfile do
 
   # Configuration for the OTP application
   def application do
-    [applications: [:logger, :inflex, :plug, :poison]]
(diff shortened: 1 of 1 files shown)


beerlington:
Thanks!

Issue #232, opened 2017-03-19
Conditionally include relationships


lleger:
I have a question about JaSerializer and relationships/including. My understanding of the spec means we can use `include` to optionally include relationships. 

Consider a serializer with this relationship:

```elixir
# post.ex
has_many :comments,
  serializer: MyApp.CommentView,
  include: false,
  identifiers: :when_included
```

If we make a request without the include parameter, this will render a response that looks something like this:

```json
{
  "jsonapi": {
    "version": "1.0"
  },
  "data": {
    "type": "post",
    "relationships": {
      "comments": {},
    },
    "id": "174f3a81-5151-4bf1-8d8d-164d59d25375",
    "attributes": {
      "updated-at": "2017-03-17T23:39:20.205508",
      "inserted-at": "2017-03-15T03:48:41.774795",
    }
  }
}
```

This isn't valid JSON API though:

> Resource linkage MUST be represented as one of the following:
> - null for empty to-one relationships.
> - an empty array ([]) for empty to-many relationships.
> - a single resource identifier object for non-empty to-one relationships.
> - an array of resource identifier objects for non-empty to-many relationships.

I _think_ the best solution would be to just remove `relationships` completely (since it's optional), in this case. Setting it to `null` or an empty array doesn't make sense to me, semantically (the relationship isn't empty, we just aren't including it in this response).

Does this make sense or am I misreading the spec? (It is also possible this option already exists in JaSerializer and I missed it.)

yogipatel:
I have been wondering the same thing, and just came across it as an issue with one of our front-end clients. Unless this is a regression, I'm surprised more people haven't run into this.

alanpeabody:
Totally agree, if the relationship has nothing it it lets remove it completely. A PR would be very welcome!

yogipatel:
@alanpeabody Sure, I'll try to check it out in the next few days.

darkbaby123:
I think it's not the best solution. The `comments` is actually a relationship object but not a resource linkage. And according to JSON API spec:

> A “relationship object” MUST contain at least one of the following:
> 
> - links: a links object containing at least one of the following:
>   - self: a link for the relationship itself (a “relationship link”). This link allows the client to directly manipulate the relationship. For example, removing an author through an article’s relationship URL would disconnect the person from the article without deleting the people resource itself. When fetched successfully, this link returns the linkage for the related resources as its primary data. (See Fetching Relationships.)
>   - related: a related resource link
[...]

PR #311, opened 2019-04-01
Don't render empty relationships

spatrik:
This PR is a fix for #232, based on #235 but based on a new master and with an added test.

With this change, a relationship with no information will not be rendered


beerlington:
Thanks!

spatrik:
Wow, that was fast :) Any plan for a new release?

PR #316, opened 2019-05-02
Implement camelcased option recommended by JSON:API 1.1

### LINKED ISSUE #315: Offer camelCase keys
Since JSON:API 1.1 will suggest to use camelCase variant for keys and it matches general JavaScript conventions it would be awesome to offer camelCase keys in `ja_serializer` (even if not the default).

strzibny:
I implemented `camelcased` option as it's very common for JSON APIs to follow JS conventions. Plus JSON:API 1.1 will recommend this as well.

Fixes #315


lib/ja_serializer/formatter/utils.ex
```diff
@@ -60,6 +60,14 @@ defmodule JaSerializer.Formatter.Utils do
   @doc false
   def do_format_key(key, :underscored), do: key
   def do_format_key(key, :dasherized), do: String.replace(key, "_", "-")
+
+  def do_format_key(key, :camelcased) do
+    camelized = Macro.camelize(key)
```
beerlington: I wasn't familiar with this so I looked it up in the docs. The docs recommend not using this for camelizing strings since it [doesn't support unicode or characters that are not valid Elixir identifiers](https://hexdocs.pm/elixir/Macro.html#camelize/1). I'm not sure what's considered a valid Elixir identifier, but unicode should probably be supported. The JSON spec says we SHOULD NOT use non-ascii chars, but it's not a "MUST NOT" so technically unicode chars are still valid. Would you be ok changing this to something that supports any unicode chars?

README.md
```diff
@@ -397,10 +397,16 @@ render conn, data: data, opts: [meta: meta_data]
 
 ### Key Format (for Attribute, Relationship and Query Param)
 
-By default keys are `dash-erized` as per the jsonapi.org recommendation, but
-keys can be customized via config.
+By default keys are `dash-erized` as per the JSON:API 1.0 recommendation, but keys can be customized via config.
 
-In your `config.exs` file:
+In your `config.exs` file you can use `camelcased` recommended by upcomming JSON:API 1.1:
```
beerlington: From what I can tell "camel case" is two words, so I think this would be better as `camel_cased`.

lib/ja_serializer/param_parser.ex
```diff
@@ -37,11 +37,16 @@ defmodule JaSerializer.ParamParser.Utils do
...
       :underscored -> key
       {:custom, module, _, fun} -> apply(module, fun, [key])
       _ -> key
     end
   end
 
   defp dash_to_underscore(key), do: String.replace(key, "-", "_")
+
+  defp uppercase_to_underscore(key) do
+    Macro.underscore(key)
```
beerlington: This is another one where the docs don't recommend using it for general underscore purposes. I would feel more comfortable if we did not use here for that reason.

README.md
```diff
@@ -397,10 +397,16 @@ render conn, data: data, opts: [meta: meta_data]
...
+
+```elixir
+config :ja_serializer,
+  key_format: :camelcased
+```
+
+Or `underscored` to keep the keys same as Elixir conventions:
```
beerlington: I'm not sure I'd say underscored attribute keys are really an Elixir convention since it's really the client that would care about this. I would just cut this to "Or `underscored`: "

README.md
strzibny: Yep, wasn't sure about it.

README.md
strzibny: Yep, no problem. But naming in Elixir really does follow underscored for keys IMHO.

lib/ja_serializer/formatter/utils.ex
strzibny: I missed it, I will check what I can replace this with.

strzibny: I feel we can use either https://github.com/nurugger07/inflex or https://github.com/sobolevn/recase. If the extra dependency is problematic, we could then merge the code with attribution.

strzibny: Ah just noticed there is Inflex already, great!

strzibny:
@beerlington Thanks for the review. I updated the code. I used a newer version of inflex simply because it fixes the bug with pascalCase. If the dependency upgrade is a no-go I can update the code with a workaround.

beerlington:
Thanks!

Issue #138, opened 2016-07-15
Chrome returning a 406 Not Acceptable response


drewrawitz:
Hello,

I started a brand new phoenix project and followed these docs but I keep running into a weird issue.  Chrome is returning a 406 error:

![](https://api.monosnap.com/rpc/file/download?id=xaVWRjQtP9tRe3h6p1VeHQaHGXhLmB)

However, if I run that same URL into Postman, I get a valid response!

![](https://api.monosnap.com/rpc/file/download?id=4jidpKLpb6w31lu9t3IUbEBn16bind)

Have you seen this before?  The steps I took were:

* Create a brand new phoenix project
* Install JaSerializer 0.9.0
* Added the following to `config.exs`
```
config :phoenix, :format_encoders,
  "json-api": Poison

config :plug, :mimes, %{
  "application/vnd.api+json" => ["json-api"]
}
```
* Added json-api to my plug pipeline
```
pipeline :api do
  plug :accepts, ["json-api"]
  plug JaSerializer.ContentTypeNegotiation
  plug JaSerializer.Deserializer
end
```
* Used the JSON API Generator
```
mix ja_serializer.gen.phoenix_api User users first_name:string last_name:string username:string email:string bio:text
```

I think that was it, did I miss anything? It's weird that it works fine in Postman.  Sorry if I missed out any important details, let me know if you need me to specify anything else.

Thank you!

alanpeabody:
Are you hitting this url directly in chrome? Or via an ajax request?

My guess is just that it doesn't have an acceptable "Accept" header. `*/*`, `application/*`, and `application/vnd.api+json` should work. Postman is probably sending one of those where as chrome is not.

Going to close this, but if the problem persists or you believe this is a bug please don't hesitate to reopen this!

drewrawitz:
@alanpeabody Thanks for your quick reply. I'm hitting the URL directly in chrome.  That should be fine, right?  

Is there anything I can do in the code to make it work?

alanpeabody:
My understanding of the spec is that this behavior is correct, but I might be mis-interpreting it:

http://jsonapi.org/format/#content-negotiation

Basically your browser is not sending the right "Accept" header for JSON-API. You could remove the ContentTypeNegotiation plug, but you would probably end up running into other problems or reproducing it. You could also use an api tool that sends an acceptable Accept header, such as postman, curl, etc.

drewrawitz:
[...]

PR #320, opened 2019-05-28
Allow accept header with quality param

lasseebert:
Related to #138

When requesting with an Accept header with a quality param, e.g. `*/*;q=0.9`, a 406 would be returned.

The quality should be used to prioritize the caller's different acceptable response content-types, but since we only care about jsonapi content type, we should just ignore it.

~~I also removed a test for Accept header `application/vnd.api+json; charset=utf-8`, since `charset` is not a valid param to Accept headers.~~

I found this when requesting a ja_serializer endpoint with Chrome. Chrome sends by default this accept header:

```
text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3
```

Which should be accepted by ja_serializer because of the `*/*;q=0.8` part.

~~(The changes looks a bit weird in the github diff editor since I removed one test and added another. [This](https://github.com/lasseebert/ja_serializer/blob/f8cded6d667a8b59467df10e52e6b5e6a8390c20/test/ja_serializer/content_type_negotiation_test.exs#L90-L99) is the test I added)~~

### DIFF
--- a/lib/ja_serializer/content_type_negotiation.ex
+++ b/lib/ja_serializer/content_type_negotiation.ex
@@ -45,6 +45,7 @@ defmodule JaSerializer.ContentTypeNegotiation do
       conn
       |> get_req_header("accept")
       |> Enum.flat_map(&String.split(&1, ","))
+      |> Enum.map(&(&1 |> String.split(";") |> hd()))
       |> Enum.map(&string_trim/1)
 
     cond do
--- a/test/ja_serializer/content_type_negotiation_test.exs
+++ b/test/ja_serializer/content_type_negotiation_test.exs
@@ -78,10 +78,10 @@ defmodule JaSerializer.ContentTypeNegotiationTest do
     assert result.status == 415
   end
 
-  test "Returns 406 Not Acceptable if all media type params" do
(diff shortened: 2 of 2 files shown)


beerlington:
@lasseebert this makes sense. The only thing that's not clear is why you removed the `406` response test. Don't we still want that?

lasseebert:
@beerlington The Accept header [only accepts the `q` param](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept), so my take was that we can throw away other params.

We can also throw away the `q` param, since we're only interessted in if the caller accepts jsonapi or not.

The deleted test tested that it would 406 on `Accept: application/vnd.api+json; charset=utf-8`, but IMHO we should just discard any invalid params. And so the `application/vnd.api+json; charset=utf-8` with discarded params becomes `application/vnd.api+json`, which is valid.

lasseebert:
Ah, I think I missed your point. We should still test 406 response and I deleted the only test that did that. Let me fix it. I'll push in a minute.

beerlington:
Thanks!

lasseebert:
Thanks for the quick review 👍

PR #336, opened 2020-10-07
Allows import to be used as a relationship name

paulanthonywilson:
If one of your entities has a relationship called "import", and you are trying to upgrade `ja_serializer` then compilation fails with something like

```
invalid argument for import, expected a compile time atom or alias, got: struct
```

This is a fix for that.


beerlington:
Thanks!

PR #348, opened 2024-06-13
fix warnings on Elixir 1.17

SteffenDE:


test/ja_serializer/phoenix_view_test.exs
```diff
@@ -130,7 +130,7 @@ defmodule JaSerializer.PhoenixViewTest do
 
   test "render conn, show.json-api, data: nil" do
     json = @view.render("show.json-api", conn: %{}, data: nil)
-    assert json['data'] == nil
+    assert json[~c"data"] == nil
```
beerlington, line 133: I think this was a typo and it should actually be `json["data"]`.

SteffenDE: good point, that does make sense - I should have looked more at the code instead of simply replacing it 😃

beerlington:
Thanks!

alejandrodevs:
@beerlington Do you plan to release a new version with the fix anytime soon?

beerlington:
@alejandrodevs thanks for the reminder! that is now released in 0.18.1.

alejandrodevs:
@beerlington Thank you so much!

